PR #20, opened 2020-07-10
perf: use static arrays for holding lookup data

### LINKED ISSUE #18: Change lookup init to use static arrays
When investigating performance/crashing issues with En-ROADS on iPad, I discovered that a large part of the memory pressure is coming from the way the lookup initialization code gets compiled down to wasm.

The current approach passes a list of (x,y) pairs as varargs to `__new_lookup`, which then copies them into a freshly malloc'd array in the `Lookup`.  Something about the way Emscripten handles the varargs seems to cause higher memory usage when the module is loaded and also increases the on-disk size of the wasm file.

An example init line currently looks like this:
```
__lookup1 = __new_lookup(111, 1990.0, 7.53332, 1991.0, 7.7605, 1993.0, 8.15487, ..., 2100.0, 0.550535);
```

Some rough baseline performance stats from my ancient (2013) iPad Air:
```
wasm on disk size: 1084036 bytes
init time: 28267 ms
memory use after init: 45 MB (JS) / 253 MB (page)
memory use after running model: 45 MB (JS) / 450 MB (page)
```

(The memory usage stats here and below are fuzzy, representing the rough max values reported using Safari developer tools.)

I tried a couple alternatives:

1. Make `__new_lookup` take an explicit `double *`, then pass a literal array, for example:
```
__lookup1 = __new_lookup(111, (double[]){ 1990.0, 7.53332, 1991.0, 7.7605, 1993.0, 8.15487, ..., 2100.0, 0.550535 });
```

This made a big improvement:
```
wasm on disk size: 775426 bytes
init time: 4469 ms
memory use after init: 81 MB (JS) / 124 MB (page)
```

2. Make `__new_lookup` take an explicit `double *`, then pass a static array declared earlier, for example:
```
const double __lookup1_data[222] = { 1990.0, 7.53332, 1991.0, 7.7605, 1993.0, 8.15487, ..., 2100.0, 0.550535 };
...
__lookup1 = __new_lookup(111, /*copy=*/false, __lookup1_data);
```

This made another big dent:
```
wasm on disk size: 773968 bytes
init time: 4046 ms
memory use after init: 64 MB (JS) / 103 MB (page)
```

chrispcampbell:
See #18 for more information and performance numbers.

Fixes #18

### DIFF
--- a/src/c/vensim.c
+++ b/src/c/vensim.c
@@ -259,7 +259,7 @@ double* _VECTOR_SORT_ORDER(double* vector, size_t size, double direction) {
     // TODO signal error
     return NULL;
   }
-  for (size_t i = 0; i < size; i++) {
+  for (int i = 0; i < size; i++) {
     d[i].x = vector[i];
     d[i].ind = i;
   }
--- a/src/c/vensim.c
+++ b/src/c/vensim.c
@@ -124,24 +124,27 @@ double _ZIDZ(double a, double b) {
 //
 // Lookups
 //
-Lookup* __new_lookup(size_t size, ...) {
+Lookup* __new_lookup(size_t size, bool copy, double *data) {
 	// Make a new Lookup with "size" number of pairs given in x, y order in a flattened list.
 	Lookup* lookup = malloc(sizeof(Lookup));
-	lookup->data = malloc(sizeof(double) * 2 * size);
 	lookup->n = size;
 	lookup->inverted_data = NULL;
-	// Copy arguments into the lookup data.
-	va_list args;
-	va_start(args, size);
-	for (size_t i = 0; i < 2 * size; i++) {
-		*(lookup->data + i) = va_arg(args, double);
+	lookup->data_is_owned = copy;
+	if (copy) {
+		// Copy array into the lookup data.
+		lookup->data = malloc(sizeof(double) * 2 * size);
+		for (size_t i = 0; i < 2 * size; i++) {
+			memcpy(lookup->data, data, size * 2 * sizeof(double));
+		}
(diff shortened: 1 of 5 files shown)


src/c/vensim.c
```diff
@@ -124,24 +124,27 @@ double _ZIDZ(double a, double b) {
...
-	// Copy arguments into the lookup data.
-	va_list args;
-	va_start(args, size);
-	for (size_t i = 0; i < 2 * size; i++) {
-		*(lookup->data + i) = va_arg(args, double);
+	lookup->data_is_owned = copy;
+	if (copy) {
+		// Copy array into the lookup data.
+		lookup->data = malloc(sizeof(double) * 2 * size);
+		for (size_t i = 0; i < 2 * size; i++) {
```
travisfranck, line 136: Why do you use size_t in this for loop but int on Line 262?    Also, I'm not 100%, but I believe at one point WASM handle some number types (double vs float, int vs ???) better than others.  This might be from WebAssembly days and not apply to WASM now.  Emscripten has evolved.

chrispcampbell: In most cases where you're indexing into an array, it's a good idea to use `size_t`, since it's an unsigned type and it matches the address size of the CPU which is important for addressing large data sets.  We use `size_t` nearly everywhere, but the loop at 262 is an anomaly because it's filling an array with index values of type `int`.  I changed that line because the compiler flagged it as a type mismatch warning; could've also fixed it by casting or changing the type of the `ind` field in `dbl_ind` to `size_t`, but figured changing the loop variable to `int` best reflects its usage.

PR #26, opened 2020-07-10
perf: replace dimension array access with simple index

### LINKED ISSUE #23: Replace dimension array access with simple index
SDEverywhere generates arrays of index numbers to help support subdimensions:

```
  function arrayDimensionsSection() {
    // Emit a declaration for each array dimension's index numbers.
    // These index number arrays will be used to indirectly reference array elements.
    // The indirection is required to support subdimensions that are a non-contiguous subset of the array elements.
    let a = R.map(dim => `const size_t ${dim.name}[${dim.size}] = { ${indexNumberList(sub(dim.name).value)} };`)
    let arrayDims = R.pipe(a, asort, lines)
    return arrayDims(allDimensions())
  }
```

These get declared at the top of the generated c file, for example:

```
// Array dimensions
const size_t _all_things[5] = { 0, 1, 2, 3, 4 };
const size_t _all_things_except_foo_and_bar[3] = { 1, 2, 4 };
```

When accessing values, these arrays are used as a form of indirection, for example:

```
for (i = 0; ...) {
  _something[_all_things[i]] = _another[_all_things[i]];
}
```

In En-ROADS, some subsets are non-contiguous, but in a majority of cases, these dimension index arrays are "trivial", meaning their elements are identical to their index/position.  For those cases, we can generate more optimal code that skips the array access; the example above would become:

```
for (i = 0; ...) {
  _something[i] = _another[i];
}
```

Avoiding unnecessary array accesses leads to some more modest performance gains.

chrispcampbell:
More details in #23.

Fixes #23

### DIFF
--- a/src/EquationGen.js
+++ b/src/EquationGen.js
@@ -11,6 +11,7 @@ const {
   hasMapping,
   isDimension,
   isIndex,
+  isTrivialDimension,
   normalizeSubscripts,
   separatedVariableIndex,
   sub
@@ -177,7 +178,12 @@ module.exports = class EquationGen extends ModelReader {
     return R.map(subscript => {
       if (isDimension(subscript)) {
         let i = this.loopIndexVars.index(subscript)
-        return `_${subscript}_${i}_`
+        if (isTrivialDimension(subscript)) {
+          // When the dimension is trivial, we can simply emit e.g. `[i]` instead of `[_dim[i]]`
+          return `_${i}_`
+        } else {
+          return `_${subscript}_${i}_`
+        }
       } else {
         return `_${sub(subscript).value}_`
       }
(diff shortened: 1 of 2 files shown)


chrispcampbell:
> If the output is the same as Vensim, this is a nice perf boost. I wonder if Todd did this because of subscript mapping, which maybe we don't use on some of these "Trivial" subscripts. Something to watch out for if model results aren't the same as Vensim.

Just to follow up on this, for each of these perf fixes, I ran both the SDEverywhere `modeltests` and the new `check` script in En-ROADS (which compares the latest model outputs against the baseline outputs from before the perf work) and haven't found any discrepancies.  Will be good to get these checks automated per our other discussions on continuous integration.

PR #37, opened 2020-09-12
fix: add async and await to some chained cli functions

### LINKED ISSUE #35: `sde test` fails to build model when a spec file is provided
Some of the `sde` CLI files, such as `sde-build.js`, have an async handler function, which works fine when that command is called directly (e.g. `sde build --spec spec.json`) but other commands are implemented by chaining and can fail when a spec file is passed on the command line.  The spec file is read in an async function, but some of the chaining cases do not use async/await.

For example:
* `sde test` calls `run` (which is currently not marked `async`)
* `run` then calls `build` (also not marked `async`)
* `build` calls `generate` (which is marked `async`)

Long story short, we need to add `async` and `await` in a couple more places to ensure that the spec file is read fully before proceeding with each step.

chrispcampbell:
Fixes #35

### DIFF
--- a/src/sde-build.js
+++ b/src/sde-build.js
@@ -18,9 +18,9 @@ let builder = {
 let handler = argv => {
   build(argv.model, argv)
 }
-let build = (model, opts) => {
+let build = async (model, opts) => {
   opts.genc = true
-  generate(model, opts)
+  await generate(model, opts)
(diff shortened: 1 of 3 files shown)


travisfranck (commented):
reviewed.

PR #38, opened 2020-09-12
fix: fix LOOKUP FORWARD to correctly handle fractional inputs

### LINKED ISSUE #36: LOOKUP FORWARD produces incorrect results with fractional input
While investigating #33, I added some additional cases to the existing `lookup` model tests to test behavior for fractional inputs and found that the existing `__lookup` implementation produces results that don't match Vensim output, specifically in the LOOKUP FORWARD case.

chrispcampbell:
This fixes `__lookup` behavior for `Forward` mode and adds more cases to the existing `lookup` model test to make sure we've covered everything.

Fixes #36


travisfranck (commented):
I didn't understand this one.

PR #43, opened 2020-09-14
perf: cache last input and last accessed index for faster lookups

### LINKED ISSUE #34: Cache last input and accessed index for faster lookups
The `LOOKUP` function operates on (x,y) pairs where the x value is assumed to be monotonically increasing.  The way lookups are currently used, the `input` value tends to be a "marching time" value, meaning that it increases from initial time to final time, and then back to initial time and so on.

Two observations:

1. In the case of the En-ROADS model, the time step is 0.125 years, but most lookup data has a granularity of one year.  This means that much of the time, each lookup is called 8 times with the same input and same expected return value, before moving up to the next year.

2. Currently the `LOOKUP` function uses a simple linear search that always begins at zero.

We can improve the speed of lookups and address both of the above by caching the last `input` and the last-accessed index value in the `Lookup` itself.  If the current `input` value is greater than or equal to the last `input` value, then we can set our start index to the last index value.  With this approach we would generally never have to go more than one index ahead, and only once the time variable goes back to initial would we need to set the start index back to zero.

chrispcampbell:
Fixes #34

### DIFF
--- a/src/c/vensim.c
+++ b/src/c/vensim.c
@@ -74,6 +74,8 @@ Lookup* __new_lookup(size_t size, bool copy, double *data) {
 		// Store a pointer to the lookup data (assumed to be static or owned elsewhere).
 		lookup->data = data;
 	}
+	lookup->last_input = DBL_MAX;
+	lookup->last_hit_index = 0;
 	return lookup;
 }
 void __delete_lookup(Lookup* lookup) {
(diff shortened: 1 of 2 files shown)


travisfranck (commented):
Is any assumption made that LOOKUP tables are always monotonically increasing?  The code should work for inverted "U" shaped functions, as well as other relational shapes.

PR #76, opened 2021-07-09
build: update antlr4 and antlr4-vensim to latest versions and convert to ES modules

### LINKED ISSUE #75: Update antlr4 to the latest version
Updating antlr4 to version 4.9.2 required a move to Node 14+ and ES modules. The parser that antlr4 generates exports ES modules that cannot be required by a CommonJS module. Therefore, SDEverywhere modules must be converted to ES modules. A partial conversion that keeps some modules in CommonJS is possible, but it's easier and cleaner to simply convert all modules to ES.

We can keep the .js file extensions (instead of .mjs) by making all modules ES in package.json with `"type": "module"` at the top level.

The index.js entry point simply loads sde.js with `import './sde.js'`. This executes the sde module. No import name is necessary. We avoid dynamic imports in sde.js by importing all modules statically up front with names that are used later.

The sde-* command modules import npm packages with the syntax `import R from 'ramda'` and import local modules with the syntax `import Model from './Model.js'`. Sometimes it is necessary to reach into a package like this: `import parseCsv from 'csv-parse/lib/sync.js'`. Named imports are used for utility functions like this: `import { canonicalName, matchRegex } from './Helpers.js'`. These are straightforward translations from the CommonJS "require" syntax.

The sde-* command modules export a default object that is used by yargs. Many of the visitor and analysis modules export a default class. Utility modules export named functions and objects with `export let ...`. This requires looking up each name in the previous `module.exports = {...}` and adding "export" to each declaration, then removing the module.exports.

The import statements determine the format of the corresponding export: whether it's a default export or a collection of named exports.

Requires inside functions are moved to the head of the module and seemed to work fine. We keep the browserify dependency optional by dynamically importing it.

The JS files in the src/web directory that are used to make web apps are left alone. This preserves compatibility with older browsers and with the browserify packager. It was not required anyway because antlr4 is not involved.

The `__dirname` global variable is replaced with the URL constructor. It can be embedded directly in a path.join without making a separate variable like this.
```
let __dirname = path.resolve(new URL('.', import.meta.url).pathname)
```

ToddFincannon:
Fixes #75 

Upgrading antlr4 to the latest version 4.9.2 had far-ranging implications. Their move to ES modules meant that SDEverywhere had to move to ES modules too. Detailed notes in issue #75. This update requires a parallel update to the antlr4-vensim module.


chrispcampbell (approved):
Looks great.  Thanks for taking care of this, Todd.  This gets us one baby step closer to eventually converting to TypeScript 😉

I decided to use a pinned version of antlr4-vensim using a GitHub URL so that we don't have to wait for a release of antlr4-vensim.  That way we can incrementally try out the changes you have queued up in SDEverywhere and antlr4-vensim, and once we've confirmed that they're all working together, we can do a proper release of antlr4-vensim.

I built En-ROADS with this version of SDE and didn't find any issues.  Great to not have the [circular dependency warnings](https://github.com/antlr/antlr4/issues/2834#issuecomment-703337241) issue after the antlr4 upgrade.

Issue #78, opened 2021-07-10
Implement subscript aliases


ToddFincannon:
The subscript alias operator <-> allows a subscript range to be defined as an alias of another subscript range with the same indices. For example:

DimE <-> DimF
DimF: F1, F2, F3

DimE will have indices F1, F2, F3.

PR #79, opened 2021-07-10
feat: implement GET DIRECT SUBSCRIPT for CSV

### LINKED ISSUE #77: Implement GET DIRECT SUBSCRIPT
Implement the [GET DIRECT SUBSCRIPT](https://www.vensim.com/documentation/fn_get_direct_subscript.html) Vensim function for CSV files where the pathname is given in first argument. This covers the immediate use case. A fuller implementation would also read XLSX files and allow tags in the spec file like the GET DIRECT DATA implementation does.

ToddFincannon:
Fixes #77 

Implement GET DIRECT SUBSCRIPT in the subscript range reader by reading a CSV file with subscript indices named in the function


chrispcampbell (changes requested):
@ToddFincannon I got the latest from `develop` merged into this branch, and I set the antlr4-vensim dependency to use the latest from your `todd/3-get-direct-subscript` branch.  So now the build proceeds, but the directsubs test fails here:
https://github.com/climateinteractive/SDEverywhere/runs/3080874108?check_suite_focus=true#step:5:62

It looks like there needs to be some code that resolves the location of the csv file relative to the mdl file, is that right?  Can you take a look?

The rest of the fix looks good to me, so I think this will be ready to be merged once that test failure is resolved.

ToddFincannon:
Quite right — I realized that and fixed the modeltests script in the following branch #78. Maybe we need to back port that fix to the #77 branch.

chrispcampbell:
> Quite right — I realized that and fixed the modeltests script in the following branch #78. Maybe we need to back port that fix to the #77 branch.

Ah, I hadn't looked at that branch yet, but I see that change now.  Seems more like a workaround though if users have to call `sde` from a particular directory (i.e., from the same directory as the model) for the csv file path to be resolved correctly.  How hard do you think it would be to make it so the csv file is resolved relative to the mdl file (assuming that's how Vensim resolves it)?

ToddFincannon:
I did the back port of the modeltests script from #78, but now the arrays_cname model test is broken. I think it's unrelated to the change in the test script. Will investigate further next week.

chrispcampbell:
> I did the back port of the modeltests script from #78, but now the arrays_cname model test is broken. I think it's unrelated to the change in the test script. Will investigate further next week.

@ToddFincannon I'm not seeing the test failure in arrays_cname that you mention.  That said, I would still prefer to not change the test script (i.e., should back out the change in 245bc48); it doesn't seem right to have to call `sde` from a particular cwd for `GET DIRECT SUBSCRIPT` to work correctly. 

Can we instead change SDE to resolve the path of the csv file relative to the mdl file?  I sketched it out incompletely [in this gist](https://gist.github.com/chrispcampbell/35eeee2996c1630fff7f7cf5d4079ef5); would something like this work?

ToddFincannon:
I backed out the modeltests change, and will do the same for #78 and succeeding branches. I implemented the code in your gist to resolve the CSV pathname relative to the model directory. The arrays_cname test only fails starting from the antlr4-vensim #4 branch. Tests are passing then antlr4-vensim is set to the `3-get-direct-subscript` branch. I will fix that separately.

chrispcampbell (approved):
Looks good, will merge once the build finishes.

PR #80, opened 2021-07-10
feat: implement the <-> subscript alias operator

### LINKED ISSUE #78: Implement subscript aliases

ToddFincannon:
Fixes #78 

Implement the <-> subscript alias operator by adding a special case to the subscript factory function with an empty string instead of an index array. Copy the referenced subscript information once all subscripts are read,


chrispcampbell (approved):
Tests pass, looks good.

PR #82, opened 2021-07-13
feat: implement GET DIRECT DATA for CSV

### LINKED ISSUE #81: Implement GET DIRECT DATA for CSV files
Currently, [GET DIRECT DATA](https://www.vensim.com/documentation/fn_get_direct_data.html) only works with Excel XLSX files. Extend the implementation to work with CSV files as described in the docs.

ToddFincannon:
Fixes #81 

Add CSV support to GET DIRECT DATA. This required refactoring the existing XLSX support a bit to abstract away data access when converting the data to a lookup. CSV tests were added to the directdata test model.


chrispcampbell (changes requested):
The new fix for resolving relative to model dir looks good, but I just remembered that we still have the issue that we spotted yesterday where `readCsv` logs the error instead of throwing. 

I think maybe it would be best to remove the try/catch in `readCsv` and let errors propagate up.  Right now the error message suggests that the operation failed because the file doesn't exist, but it could also be that the CSV data was invalid or something, so removing the try/catch might be sufficient.

Can you either fix that as part of this branch, or open a new issue and address it as a separate PR?  Either one is fine by me, but just wanted to make sure that issue didn't get lost.

ToddFincannon:
I removed the try/catch block around the CSV parse. I tested it with a bad pathname and got the following unhandled exception error message:

`Error: ENOENT: no such file or directory, open '/Users/toddfincannon/Projects/SDEverywhere/models/directdata/x_data.csv'`

I think this is clear enough that we don't need a special error message.

chrispcampbell (approved):
Todd addressed the `readCsv` error handling issue I mentioned in my last comment, and that fix has been merged into this branch (via PR #93), so this branch is now ready to be merged.  I'll merge it to `develop` shortly.

PR #86, opened 2021-07-17
feat: implement GET DIRECT CONSTANTS for CSV

### LINKED ISSUE #83: Implement GET DIRECT CONSTANTS
[GET DIRECT CONSTANTS](https://www.vensim.com/documentation/fn_get_direct_constants.html) reads constants from an Excel or CSV file. It is similar to GET DIRECT DATA, except that it does not result in a time series. Implement it for CSV files to meet immediate needs.

ToddFincannon:
Fixes #83 

GET DIRECT CONSTANTS is implemented during code generation by reading the CSV file and generating a constant initialization for each cell value. A test model is included that covers zero, one, and two dimensions of data (all the possibilities for a CSV file in Vensim). One test covers the case where subscripts occur in a different order on the LHS than in the data.


chrispcampbell (approved):
Looks good to me.  I merged the latest from `develop` into this branch, changed the path passed to `readCsv` to be relative to the model directory (like we did for previous branches), and then verified that all tests are passing, so I think this is ready to be merged.  I'll merge it shortly.

PR #89, opened 2021-07-23
feat: implement QUANTUM function

### LINKED ISSUE #87: Implement the QUANTUM function
Implement the [QUANTUM](https://www.vensim.com/documentation/fn_quantum.html) Vensim function, which is purely computational.

ToddFincannon:
Fixes #87 

Implement the QUANTUM function as a C macro with tests taken from the Vensim [help documentation](https://www.vensim.com/documentation/fn_quantum.html).


chrispcampbell (approved):
Looks good.  I pushed a small fix to add parentheses around the macro arguments, which adds a layer of safety for when the macro args are expressions (similar to what we have for other macros).

BTW, I was on the fence about whether this should be a macro or a function; when I worked on reimplementing some things as macros vs functions last year, in cases where an arg appears multiple times on the right side, I sometimes chose to implement it as a C function to avoid redundant calculations.  But in the case of `QUANTUM`, only `b` appears multiple times and (I'm assuming) it's likely to be a constant value rather than an expression, so I'm less worried about it.  And either way, we're getting into micro-optimization territory, so I'm really not too worried about it.

PR #92, opened 2021-07-24
fix: use correct subdimension index for delay aux vars

### LINKED ISSUE #91: NaN values in output in the delay test model
The delay model appears to run correctly as reported by `sde compare`, however, the D6[A3], D9[A3], and D12[A3] vars are coming up NaN at all time steps. These vars all have the second subscript of a subdimension.

ToddFincannon:
Fixes #91 

The delay test generated NaN values for aux vars subscripted with a subdimension. Aux vars are generated as part of internal delay equation generation. They are added in the analysis stage after the model vars have already been read. The problem was that vars on a subdimension must be separated into individually indexed vars, and this was not being done for the generated aux vars. There is a mechanism to add "non a-to-a" (separated vars in XMILE terminology) after the initial scan of model variables. This was used to add separated aux vars and get the correct refid for references. The DELAY1 and DELAY3 functions both needed this change in somewhat different variations. I regenerated the variable listing too to account for the new aux vars. The delay test now passes without the NaN error messages.

### DIFF
--- a/models/delay/delay_vars.txt
+++ b/models/delay/delay_vars.txt
@@ -1,301 +1,631 @@
 _aux1: aux
 = delay
 refId(__aux1)
+hasInitValue(false)
+refs(_delay)
 
-_aux10: aux
-= ((delay)/3)
+_aux10[DimA]: aux
+= delay
 refId(__aux10)
+families(_dima)
+subscripts(_dima)
+hasInitValue(false)
+refs(_delay)
 
 _aux11[DimA]: aux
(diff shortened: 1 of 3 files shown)


chrispcampbell (approved):
I don't have a full understanding of how (sub)dimensions are handled in SDE, so I will trust that this is the correct fix.  I do see that the tests now pass without the messages that were seen earlier (noted in the issue), so it looks like this is fixed.  Will merge it shortly.

PR #95, opened 2021-07-27
feat: implement NPV function

### LINKED ISSUE #94: Implement NPV
Implement the [NPV](https://www.vensim.com/documentation/fn_npv.html) Vensim function.

ToddFincannon:
Fixes #94 

This was a straightforward adaptation of the code for the TREND function with the added wrinkle of needing to generate two level vars instead of just one. This shows up when we add reference, but is otherwise unremarkable. The code for the NPV function is taken from the Vensim docs. I committed a test model that validates the expansion in the docs, but removed that in the final version, since we don't need it to validate SDEverywhere code generation.


chrispcampbell (approved):
Looks good to me.  I verified that the new `npv` test passes as expected.  I'll merge this to `develop` shortly.

PR #97, opened 2021-07-28
refactor: use prettier to enforce JavaScript and Markdown formatting

### LINKED ISSUE #96: Set standard JavaScript formatting
We have been using Prettier to format JavaScript, but without a standard for the format settings. Make this part of the project as follows (via Chris):

- update prettier to latest version in package.json
- add .prettierrc
- add “prettier:check” and “prettier:fix” scripts to package.json
- update .github/workflows/build.yaml
- format all JS files to the new standard

This will allow us to code faster by neglecting formatting as we edit files and cleaning it up later. Formatting all files now will ensure that when we format them in the future, we won't cause formatting-only changes in the git diff that are irrelevant to evaluating proposed changes.

ToddFincannon:
Fixes #96 

I made all the suggested changes and formatted all files in the `src` directory. It formatted our generated files and some minified lib files in the `src/web` directory, so I added a `.prettierignore` file to disable that. Markdown files and JSON spec files also got reformatted. I ran the model tests afterwards and all still pass.


ToddFincannon:
Looks good — are we ready to merge this?

models/arrays_varname/arrays_varname_spec.json
```diff
@@ -1,18 +1,5 @@
...
-    "a[A1]",
-    "a[A2]",
-    "a[A3]",
-    "f[A1,B1]",
-    "f[A1,B2]",
-    "f[A1,B3]",
-    "h"
-  ]
+  "inputVarNames": ["inputA[A1]", "inputA[A2]", "inputA[A3]"],
+  "outputVarNames": ["Time", "a[A1]", "a[A2]", "a[A3]", "f[A1,B1]", "f[A1,B2]", "f[A1,B3]", "h"]
```
chrispcampbell, line 4: Ugh, this kind of change is my main gripe with Prettier, and there's no way around it, so I'll just accept it like I've accepted so many other Prettier changes 😄

notes/SDEverywhere Web App Configuration Reference.md
```diff
@@ -14,36 +14,35 @@ SDEverywhere uses CSV files to specify the configuration of web apps that it gen
...
-initialView | string | first level view title \> second level view title
-trackSliders | boolean | set true to run the model as sliders are moved; set false to wait until the slider is released to run the model
-initialTime | number | start time in model runs
-startTime | number | start time to show on graphs
-endTime | number | final time in model runs
-externalDatfiles | string | optional semicolon-delimited list of Vensim DAT filenames giving external data variables
-chartDatfiles | string list | optional semicolon-delimited list of Vensim DAT filenames with chart data
-logo | string | optional filename of an app logo image
-helpUrl | string | optional URL of an online help file
+| Column           | Type        | Description                                                                                                     |
```
chrispcampbell, line 17: This is kind of an excessive change by Prettier; I like using terse table syntax, but oh well, I guess for cases like this where the text isn't changed often, it's not a big deal.

src/CodeGen.js
```diff
@@ -98,17 +98,17 @@ void initLookups() {
   function emitInitConstantsCode() {
     mode = 'init-constants'
     return `
-${chunkedFunctions('initConstants', Model.constVars(),
-'  // Initialize constants.',
-'  initLookups();'
-)}
+${chunkedFunctions('initConstants', Model.constVars(), '  // Initialize constants.', '  initLookups();')}
```
chrispcampbell, line 101: This would probably read better if we assigned the `chunkedFunctions` to a variable first and try to keep the args on multiple lines, but whatever.

chrispcampbell (approved):
As is always the case with Prettier, there are some questionable formatting changes (that I noted in some inline comments), but there's nothing to be done (or can be done) about those so that's mostly just me venting 😄   Otherwise, the changes look good and I'm glad to have this wired up.  I'll merge it to `develop` shortly.

PR #99, opened 2021-07-29
feat: implement GET DIRECT LOOKUPS including a test model

### LINKED ISSUE #98: Implement GET DIRECT LOOKUPS
Add the [GET DIRECT LOOKUPS](https://www.vensim.com/documentation/fn_get_direct_lookups.html) Vensim function. The semantics for this should be the same as for GET DIRECT DATA, which generates lookups in SDEverywhere.

ToddFincannon:
Fixes #98 

GET DIRECT LOOKUPS is treated as the equivalent of GET DIRECT DATA, which generates lookups in SDEverywhere. A test model that reads CSV data with time extending horizontally is included.


chrispcampbell (approved):
Looks good to me.

PR #101, opened 2021-08-04
fix: correct initialization of 2D arrays to allow dimensions with matching subscript names

### LINKED ISSUE #84: 2D array initialization is incorrect when dimensions have matching subscript names
The C-ROADS model uses a few 2D arrays (constant lists) where the dimensions have matching subscript names, but SDEverywhere does not initialize the array values correctly.  Here is a modification to the existing `arrays_cname` test that demonstrates the issue and causes a test failure:

```
DimC: C1, C2, C3 ~~|
DimC': DimC ~~|

sc[DimC,DimC']=
	11,12,13;
	21,22,23;
	31,32,33;
	~
	~	2D constant array where dimensions have matching subscript names
	|
```

The problem is that in SDE, the code that visits 2D dimensions does lookups based on subscript names, but that breaks down when the "left" and "right" subscripts have the same names.

chrispcampbell:
Fixes #84 

This fixes an issue that we encountered when trying a newer version of the C-ROADS model.  See issue for more details.

The previous code would generate incorrect index values in the loop (for example, `[i][i]` instead of `[i][j]`) because it would look up the index of the subscript by name (and this would fail if the same subscript name appears in both dimensions).  The new code looks up the index by position in the dimensions array rather than by name.

I updated the existing `arrays_cname` test case with an example that caused SDE to fail, but now passes with the fix in place.


ToddFincannon (approved):
I'm surprised this Vensim syntax works, but clearly it does. Good to have it covered.

chrispcampbell:
> I'm surprised this Vensim syntax works, but clearly it does. Good to have it covered.

Yeah, there are a few places in the En-ROADS model where this sort of alias was used, so I guess it works in Vensim, and apparently also does in SDE though I didn't take the time to figure out how.  I was tempted to change it to use the new alias syntax you added support for recently, but figured I'd leave it as is so that there's some amount of test coverage.

PR #103, opened 2021-08-05
perf: optimize IF THEN ELSE for cases where condition expression resolves to a constant

### LINKED ISSUE #102: Optimize IF THEN ELSE for cases where condition expression resolves to a constant
In the En-ROADS and C-ROADS models, we have a number of equations that are gated by a kind of "feature flag" for things that aren't used in the app but are made available for internal testing and calibration by modelers.  The construct usually has the following form:

```
  a = IF THEN ELSE(Choose scenario > 0, Something complex, Something simple) ~~|
```

Currently there is no way for SDE to optimize these out when the conditional resolves to a constant value, so we end up generating code (and including data) that doesn't actually get used in the actual app.

I found a way to optimize this by adding a simple `ExprReader` class that determines whether an expression tree resolves to a single constant numeric value, and then for the case where the condition is constant, `EquationReader` only needs to visit one branch (the other one can be culled by the unused reference elimination code), and `EquationGen` can generate the single branch instead of the full `IF THEN ELSE`.

This change decreases the generated WASM binary for En-ROADS by ~25%, and makes the model run ~10% faster.

chrispcampbell:
Fixes #102

See issue for more details.  I updated the existing `prune` test (which is used to verify the unused reference elimination code) with new cases to cover the `IF THEN ELSE` optimizations.

At some point I'd like to reimplement the equation reader phase to produce objects that are AST-like, and then have the code gen phase use those objects instead of revisiting the whole model structure.  That way optimizations like this and the dead code elimination (and others we could do, like constant folding) could be expressed as AST transforms.  But that's for another day (probably way down the road).  For now, I think this gets the job done, and wasn't too terribly invasive.


ToddFincannon (approved):
That's a surprisingly effective optimization. Because `ExprReader` is specifically oriented to analyzing constants, I might have named it something less general. But it also makes sense to keep expanding its capability as necessary under the more general name.

chrispcampbell:
> That's a surprisingly effective optimization. Because `ExprReader` is specifically oriented to analyzing constants, I might have named it something less general. But it also makes sense to keep expanding its capability as necessary under the more general name.

That was my thinking as well, so I left it with the more general name.  Easy to change later if we find a need for different kinds of expression readers.

PR #108, opened 2021-09-11
feat: implement DELAY FIXED function

### LINKED ISSUE #29: Support for DELAY FIXED
Support for the `DELAY FIXED` function would be very useful. Does not seem like there's a way to simulate that function using `DELAY1`, though if there is, that'd be equally good.

ToddFincannon:
Fixes #29 

`DELAY FIXED` requires something none of the other Vensim functions implemented so far need: auxiliary storage of past variable values. The number of values to be stored depends on the delay time, which is quantized to a multiple of TIME STEP. The storage buffer must be allocated at runtime, because the delay time is not known until then. To support this, a new `FixedDelay` structure is added that implements a ring buffer sized to the number of delay steps.

I followed the Vensim docs which mention that `DELAY FIXED` is a level var with a special fixed delay subtype. It behaves very much like a level, with a separate initial value set in the init phase, and with evaluation at the same time as other levels. The subtype indicates the special handling that is required, starting with the `FixedDelay` support structure mentioned above. Code is generated to allocate and initialize this structure at init time just after setting the initial value for the `DELAY FIXED` variable. Note that the `FixedDelay` structure is *not* a Vensim variable. It is generated as a special case.

The other special handling that is needed is: rewriting the `DELAY FIXED` equation at eval time to remove the initialization arguments (similar to INTEG) and add the `FixedDelay` support argument; and declaring the `FixedDelay` pointer at decl time.

When the `DELAY FIXED` variable is subscripted, the `FixedDelay` struct is also subscripted. It uses the LHS subscripts. Other variables occurring in arguments on the RHS are still free to use different subscripts.

A new `_DELAY_FIXED` Vensim function in C rounds out the implementation. It is coupled with `__new_fixed_delay` to allocate and initialize the `FixedDelay` structure. `_DELAY_FIXED` implements a straightforward ring buffer with one twist: when pulling a delayed value from the buffer, it increments the current position first. This is because `DELAY FIXED` is a level with a value evaluated for the next time step.


src/c/vensim.c
```diff
@@ -404,3 +403,37 @@ double* _ALLOCATE_AVAILABLE(
...
+  // Cache input values in a ring buffer for the number of time steps equal to the delay time.
+  // Return the init value until the time reaches the delay time.
+  double result = 0.0;
+  // Require the buffer size to be positive to protect from buffer overflows.
+  if (fixed_delay->n > 0) {
+    fixed_delay->data[fixed_delay->data_index] = input;
+    // Because DELAY FIXED is a level, get the value one time step ahead in the buffer.
+    fixed_delay->data_index = (fixed_delay->data_index + 1) % fixed_delay->n;
+    // Compare the current time to the delay time up to an epsilon difference.
+    if (_time < fixed_delay->n * _time_step - 1e-6) {
```
chrispcampbell, line 432: Hmm, correct me if I'm wrong, but isn't the delay time a delta relative to the `INITIAL TIME` of the model?  If so, I think this line only works if `INITIAL TIME` is 0 (as is the case in the test model) and would instead need to be something like:
```
if (_time < _initial_time + fixed_delay->n * _time_step - 1e-6) {
```

If this is the right fix, I'd suggest changing the `delayfixed` test model to use a non-zero `INITIAL TIME` so that we cover that case.

chrispcampbell (changes requested):
@ToddFincannon Only spotted one potential issue that I noted in a comment.  The rest looks great to me.  Once you resolve that issue, this will be ready to merge.

ToddFincannon:
The suggested change worked after an additional fix. Because DELAY FIXED makes a level variable, when we check to see whether we have reached the delay time yet, we have to compare TIME to the *following* time step. This was true before, but the issue was masked by the fact that the initial value and early buffer values were both zero.

I added a new test model called `delayfixed2` with a different INITIAL TIME and two different delay times.

src/c/vensim.c
chrispcampbell: Todd addressed this properly (and added a separate test) in 4d9fe70.

chrispcampbell (approved):
The new fixes and tests look good.  I'll merge this shortly.

Issue #113, opened 2021-09-16
Direct data read fails when data dimension is not the first subscript


ToddFincannon:
Direct data variables read from an Excel or CSV file may have a dimension on the LHS. In this case, the data is a 2D matrix with time on one axis and the dimension indices on the other axis. We handled this before, but assumed that there would be just one subscript on the LHS, and took the first one as the dimension to apply the data table to. But if there is more than one subscript, this can fail if the subscript is not the first one. For instance, in this model:

```
DimA: A1, A2 ~~|
DimB: B1, B2 ~~|
i[A1, DimB] = GET DIRECT DATA('e_data.csv', ',', 'A', 'B2') ~~|
j[A1, DimB] = i[A1, DimB] ~~|

FINAL TIME = 2050 ~~|
INITIAL TIME = 1990 ~~|
TIME STEP = 1 ~~|
SAVEPER = TIME STEP ~~|
```

the dimension for the table data is the second subscript, and the wrong column is read from the table.

ToddFincannon:
Instead of assuming that the first subscript is the one to use for the data, consult the `separationDims` property in the variable, and then find the subscript that uses that dimension. It will be an index that was previously separated on the dimension. Use the numeric index value to offset to the correct row or column in the table.

PR #114, opened 2021-09-16
get direct data offset from the separated dimension

ToddFincannon:
See #113 for details. New case added to the directdata model.


src/EquationGen.js
```diff
@@ -388,11 +388,19 @@ export default class EquationGen extends ModelReader {
...
       if (getCellValue) {
         let indexNum = 0
-        if (!R.isEmpty(this.var.subscripts)) {
+        if (!R.isEmpty(this.var.separationDims)) {
           // Generate a lookup for a separated index in the variable's dimension.
-          // TODO allow the index to be in either position of a 2D subscript
-          let ind = sub(this.var.subscripts[0])
-          indexNum = ind.value
+          if (this.var.separationDims.length > 1) {
+            console.error(`WARNING: direct data variable ${this.var.varName} separated on more than one dimension`)
```
chrispcampbell, line 394: This is not a blocking comment because I think we have other existing cases like this (treating unexpected cases as a soft warning), but I wonder if we should make this and similar cases a hard error (exit with non-zero code).  At some point it might be good to revisit error handling in SDE to be a bit more fail-fast and consistent with error reporting mechanisms, but that's for another day.

chrispcampbell (approved):
@ToddFincannon Looks good to me.  I added one non-blocking comment about error handling but it's more a thought for another day.  I'll merge this one shortly.

PR #118, opened 2021-09-22
fix: expand variables allowing for any number of indices in EXCEPT clause

### LINKED ISSUE #117: Allow more than two indices in EXCEPT clauses
When a subscript in an EXCEPT clause is an index, it causes the variable to be separated on that dimension.

This var is separated on all three subscriptis:
```
except3[DimD, DimE, DimF] :EXCEPT: [D2, E2, F2] = 3 ~~|
```

But this one on only two:
```
except2[DimD, DimE, DimF] :EXCEPT: [D2, E2, DimF] = 2 ~~|
```

SDEverywhere currently only supports expansion on two subscripts in an except clause. The first var shown above ignores the EXCEPT clause, resulting in assignment to every index in the 3D subscript range:

```
// except3[DimD,DimE,DimF]:EXCEPT:[D2,E2,F2] = 3
for (size_t i = 0; i < 2; i++) {
  for (size_t j = 0; j < 2; j++) {
    for (size_t k = 0; k < 2; k++) {
      _except3[i][j][k] = 3.0;
    }
  }
}
```

We need to expand this to at least three subscripts, and preferably any number.

ToddFincannon:
Fixes #117.

See #117 for details.


chrispcampbell (approved):
@ToddFincannon This looks good to me; nice simplification.  The updated `except.dat` file wasn't included earlier, so I used Vensim to update it and verified that the tests pass.  Will merge it shortly.

PR #123, opened 2021-09-27
fix: make `sde log` wait for DAT file to be fully written and improve error handling

### LINKED ISSUE #122: Tests intermittently fail without causing the build to fail
I just noticed that sometimes one or more tests will fail (sporadically) when running on GitHub Actions, for example:
https://github.com/climateinteractive/SDEverywhere/runs/3677720774?check_suite_focus=true#step:4:65

Example error message:
```
Testing the initial model
../../models/initial/output/initial.dat not found
```

This has apparently been happening for a while now.  I went back through the build logs and found examples of this going back as far as this build on July 24:
https://github.com/climateinteractive/SDEverywhere/actions/runs/1063070071

I didn't find examples of the issue in builds before that one, but it doesn't necessarily mean that change is to blame (and it probably isn't).

There are two issues at play here:
1. A test is failing when it shouldn't.
2. The build is not failing when `sde test` encounters an error; the build should fail in this case.

chrispcampbell:
Fixes #122 

See issue for more details.  The primary fix (wait for the write buffer to be flushed) is in `sde-log.js`, and the rest improves the error handling for `sde test` so that it makes the build fail fast.


ToddFincannon:
That all looks reasonable. I was not able to test the intermittent write failure, since I don't have a failing test. Generating and catching an exception on the other errors instead of just printing an error message is the right way to do it.

ToddFincannon (approved):
Approving this PR with the previously posted comments

PR #125, opened 2021-10-06
fix: handle subdimensions correctly for GET DIRECT CONSTANTS

### LINKED ISSUE #124: Incorrect code gen for subdimension set by GET DIRECT CONSTANTS
Code generation is incorrect when an array is initialized by a combination of literal constants and values read from a file by `GET DIRECT CONSTANTS`. Two equations set the constant values. GET DIRECT CONSTANTS initializes elements given by a subdimension, with the remaining elements carved out by an EXCEPT clause.

```
DimA: A1, A2, A3 ~~|
SubA: A2, A3 ~~|
a[SubA] = GET DIRECT CONSTANTS('s.csv', ',', 'B2') ~~|
a[DimA] :EXCEPT: [SubA] = 0 ~~|
```

The first value from the CSV file is repeated for all elements in the subdimension:

```
// a[SubA] = GET DIRECT CONSTANTS('s.csv',',','B2')
_a[1] = 12.0;
// a[SubA] = GET DIRECT CONSTANTS('s.csv',',','B2')
_a[2] = 12.0;
```

The literal const goes to all elements of the subscript family, not just the elements remaining after the EXCEPT:

```
// a[DimA]:EXCEPT:[SubA] = 0
_a[0] = 0.0;
// a[DimA]:EXCEPT:[SubA] = 0
_a[1] = 0.0;
// a[DimA]:EXCEPT:[SubA] = 0
_a[2] = 0.0;
```

### LINKED ISSUE #134: Regression in EXCEPT handling caused by recent commit
@ToddFincannon Sorry for not catching this sooner, but apparently when I tested your fix for #117 using En-ROADS, I hadn't properly npm linked SDE, so I was using the wrong version of SDE.

The following construct in En-ROADS worked with 7c51641 (corresponds to PR #116), but started failing to compile with d92343e (corresponds to PR #118):

```
Elec Paths : ECoal,ECoal CCS,EOil,EOil CCS,EGas,EGas CCS,EBio,EBio CCS,nuclear,hydro,\
		Wind,Solar,Geothermal,Other renew,new
	~	
	~		|

Renewable types : Wind,Solar,Geothermal,Other renew
	~	
	~		|

Storage Cost[Renewable types]  = 
        Pre breakthrough cost for storage 
             * ( 1
                  - Post breakthrough effect for storage )  ~~|

Storage Cost[Elec Paths] :EXCEPT: [Renewable types] = 0
	~	$/GJ
	~	Marginal storage cost, only applicable for renewables.
	|
```

Running `sde generate --genc` produces the following error:

```
ERROR: no var found for refId : '_pre_breakthrough_cost_for_storage'

ERROR: no var found for refId : '_post_breakthrough_effect_for_storage'

TypeError: Cannot read property 'varType' of undefined
    at propEq (/Users/campbell/Projects/CI/SDEverywhere-alt3/node_modules/ramda/src/propEq.js:38:25)
    at /Users/campbell/Projects/CI/SDEverywhere-alt3/node_modules/ramda/src/internal/_curry3.js:39:18
    at f1 (/Users/campbell/Projects/CI/SDEverywhere-alt3/node_modules/ramda/src/internal/_curry1.js:19:17)
    at _filter (/Users/campbell/Projects/CI/SDEverywhere-alt3/node_modules/ramda/src/internal/_filter.js:7:9)
    at /Users/campbell/Projects/CI/SDEverywhere-alt3/node_modules/ramda/src/filter.js:69:3
    at /Users/campbell/Projects/CI/SDEverywhere-alt3/node_modules/ramda/src/internal/_dispatchable.js:50:15
    at Object.f2 [as filter] (/Users/campbell/Projects/CI/SDEverywhere-alt3/node_modules/ramda/src/internal/_curry2.js:34:14)
    at refs (file:///Users/campbell/Projects/CI/SDEverywhere-alt3/src/Model.js:724:21)
    at file:///Users/campbell/Projects/CI/SDEverywhere-alt3/src/Model.js:718:35
    at _map (/Users/campbell/Projects/CI/SDEverywhere-alt3/node_modules/ramda/src/internal/_map.js:7:19)
```

I think there are actually two bugs here, or rather there are two symptoms that might be traced back to the same bug.  I pushed some standalone tests that reproduce the issue to the [chris/134-except-regression](https://github.com/climateinteractive/SDEverywhere/compare/chris/134-except-regression) branch.

1. If you run the `except2` test, it reproduces the issue above ("no var found for refId") in a standalone test case.  This one uses a `spec.json` file to specify a subset of outputs.

2. If you run the modified `except` test, it seems to calculate `z total` incorrectly and results in differences.  This one does not use a `spec.json` file.  The generated C code in this case looks like this:

```
  // z total = SUM(z[SubA!])
  double __t1 = 0.0;
	for (size_t v = 0; v < 2; v++) {
	  __t1 += _z[_suba[v]];
	}
  _z_total = __t1;
  // z[SubA] = z ref a*z ref b
  _z[1] = _z_ref_a * _z_ref_b;
  // z[SubA] = z ref a*z ref b
  _z[2] = _z_ref_a * _z_ref_b;
```

My guess is that we're not setting some `references` when the variable is expanded so the dependency chain is not computed correctly (and dependents may be dropped during the dead code removal phase as a result), but I haven't investigated this too deeply.

ToddFincannon:
Fixes #124 
Fixes #134 

Details in issue #124.

(Edit by Chris: This also resolves the regression reported in #134, so I've included my `except` test modifications on this branch so that we have additional test coverage for that particular case, which did not use `GET DIRECT CONSTANTS`.)


chrispcampbell (approved):
@ToddFincannon I verified that both En-ROADS and C-ROADS are able to be built with this branch (it resolves the En-ROADS regression that I reported in #134).  I included the tests I wrote for that issue in this branch for additional coverage.

I'm trusting that the changes here resolve the issues at hand since the tests all pass, but I wish I had a better understanding of the subscript/dimension handling code in SDE (and how they're handled by Vensim in general).  I'm keen to (someday) tease some of the subscript handling code out into separate modules that can be more precisely unit tested, which I think would help make it easier for me to understand how these pieces fit together, but that's for another day.  In the meantime, I'll get this merged and will continue with the other PRs.

Issue #128, opened 2021-10-07
VECTOR ELM MAP with dimension name in offset generates incorrect code


ToddFincannon:
When a subscript name appears as a constant for the offset in VECTOR ELM MAP:

```
DimA: A1, A2, A3 ~~|
DimX : one, two, three, four, five ~~|
x[DimX] = 1, 2, 3, 4, 5 ~~|
y[DimA] = VECTOR ELM MAP(x[three], (DimA - 1)) ~~|
```

the offset appears in the wrong place in the generated code:

```
for (size_t i = 0; i < 3; i++) {
  _y[i] = ((i + 1) - 1.0)_x[_dimx[2 + ]];
}
```

PR #129, opened 2021-10-08
emit any expression for the offset arg of VECTOR ELM MAP

ToddFincannon:
Details in issue #128


chrispcampbell (approved):
Looks good, will merge.

PR #138, opened 2021-10-13
feat: increase the number of dimension and array loop index vars

### LINKED ISSUE #137: Increase the number of available loop index vars
Some very complex formulae in EPS use as many as 10 marked dimensions in SUM terms. Each of these needs a loop index variable. Previously, we only allocated two such variables for marked dimensions: [v, w]. We need to increase this to at least 10. At the same time, we need to prepare for subscripted variables with more than three dimensions in EPS. The loop index vars for dimensions are taken from a separate list, currently [i, j, k].

ToddFincannon:
Fixes #137


chrispcampbell (approved):
@ToddFincannon Looks good to me.  I was thinking that another option would be to maybe use a two letter name for `arrayIndexVars`, something like `ai, aj, ak, al, am`, but it's not too often that I find myself reviewing the generated loops, and the letters you chose seem fine (and are easy enough to change later).  Will merge shortly.

PR #160, opened 2021-11-04
fix: prevent memory leaks in fixed delay initialization

### LINKED ISSUE #159: Memory leak detected in generated C code
Emscripten revealed a memory leak in code that SDEverywhere generates. A WebAssembly module typically allocates all the memory for the heap at the beginning, unlike a C program requests more RAM from the OS if necessary. After a few runs, the EPS WebAssembly module would run out of memory. This was clearly rooted in a memory leak in the C code that SDEverywhere generates.

ToddFincannon:
Fixes #159 

I made one other change in `vensim.c` to make it possible to compile with gcc, which does not support UTF-8 source files like clang does. I had used clever μ and σ variable names in probability-related code. I replaced these with `mu` and `sigma` and SDEverywhere C code compiled with gcc just fine.

### DIFF
--- a/src/EquationGen.js
+++ b/src/EquationGen.js
@@ -809,9 +809,8 @@ export default class EquationGen extends ModelReader {
         exprs[i].accept(this)
         // For DELAY FIXED, also initialize the support struct out of band, as it is not a Vensim var.
         if (fn === '_DELAY_FIXED') {
-          this.emit(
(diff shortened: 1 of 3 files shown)


chrispcampbell (approved):
@ToddFincannon  Looks good.  Apologies for the delay in getting to this and other PRs.

PR #166, opened 2021-11-19
fix: remove unnecessary memcpy loop in lookup data initialization

### LINKED ISSUE #165: Lookup data copied multiple times at initialization
In `__new_lookup`, when lookup data is copied from an external buffer into the lookup data buffer, a loop copies it multiple times. This is left over from previous code that copied the data element-by-element in a loop.

ToddFincannon:
Fixes #165


chrispcampbell (approved):
@ToddFincannon Good catch, and thanks for fixing my blunder!

PR #168, opened 2021-11-19
perf: cache parsed CSV file data and replace Array with Set to improve code gen performance

### LINKED ISSUE #167: Generating C code is slow for EPS
Generating C code for the EPS model takes 108 seconds on my Mac, which seems way too slow.

```
time node src/sde.js generate --genc --spec ~/Projects/eps-sde/co2e_spec.json ~/Projects/eps-sde/eps.mdl
```

ToddFincannon:
Fixes #167


src/Helpers.js
```diff
@@ -321,16 +323,21 @@ export let readCsv = (pathname, delimiter = ',') => {
...
+    const CSV_PARSE_OPTS = {
+      delimiter,
+      columns: false,
+      trim: true,
+      skip_empty_lines: true,
+      skip_lines_with_empty_values: true
+    }
+    let data = B.read(pathname)
+    csv = parseCsv(data, CSV_PARSE_OPTS)
+    csvData.set(pathname, csv)
```
chrispcampbell, line 338: I wonder if we need to worry about caching many large files in memory being an issue for models that read many large CSV files (but that only make use of parts of them).  But I suspect the answer is "probably not", and if we do ever encounter such a beast, we can worry about it at that time.

chrispcampbell (commented):
@ToddFincannon I was just about to approve and merge this but saw your latest comment in #167.  I can either hold off on merging if you think you have other changes to make as part of this PR (i.e., make this PR a general "make EPS code gen faster" thing), or I can merge this without the "fixes" tag so that we leave the issue open and tackle it with multiple PRs.  Let me know which you prefer.

ToddFincannon:
That optimization reduced the full output genc time by 3x (see #167 for the flame graph). I think it's ready to merge now.

I worried briefly about the memory footprint of caching the CSVs, but it doesn't seem to be a problem. In EPS we cache 700 mostly small CSVs, and computers have lots of memory these days anyway.

chrispcampbell (approved):
Nice improvement!

PR #175, opened 2022-03-19
fix: use a global replace to join multiple line Vensim equations in comments

### LINKED ISSUE #174: Equations with more than two lines result in bad comments
Long Vensim equations extending over multiple lines may be turned into a bad comment. For instance,

```
This Year Change in Process Emissions by Policy and Industry[water and waste 36T39,indst methane destruction\
		,Pollutant]=
	Industry Process Emissions Policy to Pollutant Mappings[indst methane destruction,Pollutant\
		]
	*
	(
	  -SUM(Process Emissions Abated due to Carbon Tax by Marginal Cost[waste management methane destruction\
		,Marginal Abatement Cost
!])

	  - (
	      SUM(PERAC Mass CO2e Avoidable by Marginal Cost[waste management methane destruction\
		,Marginal Abatement Cost!])
	      - SUM(Process Emissions Abated due to Carbon Tax by Marginal Cost[waste management methane destruction\
		,Marginal Abatement Cost
!])
	    )
	    * Fraction of Methane Destruction Opportunities Achieved This Year[water and waste 36T39\
		]
	) ~~|
```

becomes:

```
// This Year Change in Process Emissions by Policy and Industry[water and waste 36T39,indst methane destruction,Pollutant] = Industry Process Emissions Policy to Pollutant Mappings[indst methane destruction,Pollutant]*(-SUM(Process Emissions Abated due to Carbon Tax by Marginal Cost[waste management methane destruction,Marginal Abatement Cost!])-(SUM(PERAC Mass CO2e Avoidable by Marginal Cost[waste management methane destruction,Marginal Abatement Cost!])-SUM(Process Emissions Abated due to Carbon Tax by Marginal Cost[waste management methane destruction,Marginal Abatement Cost
!]))*Fraction of Methane Destruction Opportunities Achieved This Year[water and waste 36T39])
```

Note the join fails to include the last line of text.

ToddFincannon:
Fixes #174


chrispcampbell (changes requested):
@ToddFincannon: The fix looks harmless, but could you please add a test case for this?  I tried quickly adding one myself but wasn't able to reproduce since I wasn't sure which part of that example equation is triggering the issue.  (Like, is it the length of the entire equation, or the multiple backslash continuations, or a combination?)

ToddFincannon:
@chrispcampbell: I didn't include a test case because I believed the change to be simple enough that it was not required. Any input with a single newline will give the same result that we had before. I tried extracting the long equation from EPS that triggered this bug, but it pulled in so many other things, that it would take a long time to get a valid equation. I can say that this change fixed the EPS problem in situ. I prefer not to take the effort to construct another test case.

chrispcampbell:
@ToddFincannon: OK, understood.  I just wanted to make sure we have an independent test case on hand for changes like this to ensure coverage for refactoring work I'd like to get to someday.  I was able to build an isolated test case (see `longeqns`, included in this branch) using the EPS equation you included in the issue.  The names weren't important, just the structure (and line break locations) of the equation, so it ended up being pretty quick to add.  I'll merge this PR shortly.

chrispcampbell (approved):

PR #235, opened 2022-09-22
feat: add `create` package for simpler project init

### LINKED ISSUE #228: Add `create` package for simpler project init
Currently it is not as easy as it could be to get started with SDE.  With the new build and plugin system, we have most of the foundation in place, and there is a `hello-world` project that can be used as a kind of template, but it would be better if the user could just type the following:

```
# Assume that my-dir contains a Vensim model
cd my-dir
npm create @sdeverywhere
npm start
```

Ideally that one `create` command would create files such as `package.json`, `sde.config.js`, a "core" library (that wraps the generated wasm model), and a simple app template to get the user started.

The `start` command would then start the local dev builder, which runs the generated app and opens the model-check report.

chrispcampbell:
Fixes #228 

See issue for more details.  This adds a new `@sdeverywhere/create` package that works in conjunction with `npm create` so that you can now type:

```sh
npm create @sdeverywhere
```

It walks the user through a series of steps to automate the process of setting up a new project that uses SDEverywhere, so you can go from Vensim mdl to library / app / model-check in a few seconds.

As part of this, there are two new packages, `template-default` and `template-minimal` under the `examples` directory.  These are used as templates by the `create` package.  We could possibly add others later to cover different use cases.

Here's a screenshot:

Screen Shot 2022-09-21 at 10 25 25 PM

/cc @ToddFincannon


chrispcampbell:
I'll be merging this as is so that I can publish the initial versions of the `create` and `plugin-config` packages and do some testing to make sure they work as intended.

PR #264, opened 2022-10-24
feat: implement DEPRECIATE STRAIGHTLINE

### LINKED ISSUE #258: Implement the DEPRECIATE STRAIGHTLINE function
The [DEPRECIATE STRAIGHTLINE](https://www.vensim.com/documentation/fn_depreciate_straightline.html) Vensim function is used in EPS 3.4.2, so we need to support it in SDEverywhere.

ToddFincannonEI:
Fixes #258 

This PR implements the Vensim `DEPRECIATE STRAIGHTLINE` function as required in #258.

The C implementation is similar to `FIXED DELAY`. It uses a ring buffer to accumulate contributions from an input stream uniformly spread over a depreciation time. Unlike `FIXED DELAY`, which is a special kind of level, `DEPRECIATE STRAIGHTLINE` is an aux variable. The implementation assigns a variable subtype of `depreciation` to trigger special handling in code gen. Also unlike `FIXED DELAY`, it is necessary to initialize the ring buffer values to zero, because the buffer contents will start to be emitted on the very first time step. (The `FIXED DELAY` buffer has to fill up first before emitting any data.)

A new `depreciate` test model demonstrates the function.

The `fisc` fiscal period argument is not supported in this implementation.


chrispcampbell (approved):
@ToddFincannonEI: This all looks like a clean addition to me, thanks.

Re: this comment in the PR description:

> The fisc fiscal period argument is not supported in this implementation.

I think it would be good to have a table in the README (or somewhere in the docs) that lists out all the supported Vensim functions along with any per-function limitations/caveats.  I just filed #265 today to track improvements I've been wanting to make to the README and docs in general, so assuming I add that table of supported functions as part of that work, I can include a note about fiscal period being unimplemented at this time.

PR #288, opened 2022-12-08
fix: check for valid numbers when reading a CSV file

### LINKED ISSUE #284: Check CSV data for numeric cell values
SDEverywhere assumes that all cells have numeric values. The equation generator for `GET DIRECT DATA` and `GET DIRECT CONSTANTS` throws an unhandled exception when the value is text. We discovered this when bad data crept into some CSV files. Vensim silently converts text cells in CSV files to zeroes without issuing a warning, so it was only apparent when converting the model with SDEverywhere. This needs to be handled better.

ToddFincannonEI:
Fixes #284 

The implementation of `GET DIRECT DATA` and `GET DIRECT CONSTANTS` relies on the `cdbl` function to convert a CSV value into a C double literal string. The`csv-parse` package we use to read the CSV gives a number for cells with numeric values. `cdbl` was simply using `toString` to convert the number to a string. When the cell value was a string instead, `cdbl` would fail with an unhandled exception.

The solution follows Excel practice by trying to convert a string value to a number, with blank strings converted to zero. This is done with `parseFloat`. If `NaN` is returned, `cdbl` throws an exception with a helpful error message.

The equation generator for `GET DIRECT DATA` and `GET DIRECT CONSTANTS` catches the exception and issues an error message indicating the file with the bad data. This has been helpful for tracking down previously unknown bad data that Vensim ignored.

This solution does not handle every case that might come up. The JavaScript `parseFloat` method converts inputs such as '1ABC' to the number 1. A full solution would test for every valid Excel number format, but that would be much more complicated. We might be able to find a package that solves this problem, at the cost of introducing a dependency.

I went ahead and did a major version upgrade to the latest `csv-parse` package.


chrispcampbell:
@ToddFincannonEI: Out of curiosity, what version of Vensim did you test with?  I tried updating the existing `directconst` sample model to see if we could have an automated test that covers this, but Vensim (I'm using 9.2.1) raises an error if I set a cell in a CSV file to something like `ZERO` or an empty string:

```
No number found in cell C3 (and perhaps others) of data/h.csv:,.
```

I pushed my changes to `directconst` to this branch, but we can back it out depending on the answer to the following question.

And the question is: Could we continue to treat this as a hard / fail-fast error like before, but keep your improved error messages (which are helpful for tracking down the source of the error)?  Or is there a good reason to keep it as a soft error?

I like to err on the side of failing fast for cases like this (similar to how Vensim handles it), because otherwise it's easy to overlook error messages (that aren't treated as hard errors) in the build logs.

ToddFincannonEI:
I used Mac version 9.3.5 with EPS. I didn't try to make a simple test case, having fixed the problem in EPS.

ToddFincannonEI:
The reason I kept it as a soft failure is so it would continue on to find other locations with bad data.

chrispcampbell:
> The reason I kept it as a soft failure is so it would continue on to find other locations with bad data.

OK, fair enough.  In that case, I think I'll leave my change to the `directconst` test case intact, and will merge this as is.  I filed a separate issue (#290) for considering a command-line flag to control whether errors like this are treated as a soft warning or hard error.  That way we don't hold up this PR and can consider the issue more broadly across SDE at a later time.

chrispcampbell (approved):
This looks good (taking into account the other comments I made) so I'll get it merged now.

PR #289, opened 2022-12-08
fix: reset FixedDelay state at the beginning of each run

### LINKED ISSUE #285: Reallocate the FIXED DELAY data buffer when delay time changes
The C implementation of the `FIXED DELAY` function assumes that the delay time will not change over the course of a run. It dynamically allocates a data buffer with a size determined by the delay time the first time it is called. The delay time can change in a future time step. When this happens, the buffer needs to be reallocated.

ToddFincannonEI:
Fixes #285 

The solution detects a change in the delay time and reallocates the data buffer accordingly. This required a bit of refactoring to prevent duplicate calculation of the number of buffer elements and the buffer size.

In the issue I forgot to mention that this situation only comes up when an SDEverywhere model is run more than once. Inputs to subsequent runs can cause a recalculation of the delay time argument to `FIXED DELAY`. Because the buffer is dynamically allocated, it is preserved across model runs. It needs to be reset on subsequent runs even if the buffer size has not changed. The solution handles that too.


chrispcampbell (approved):
This looks good to me.  I wish there was an automated test for this but understand that we don't have the infrastructure at the moment to have tests that verify results for multiple run iterations.  (I'm planning to add JS-level integration tests that should make that sort of thing possible, but not quite there yet.)  Anyway!  I'll merge this now.

Issue #291, opened 2022-12-09
Runtime packages fail if SAVEPER is not equal to 1


chrispcampbell:
When I first developed the `runtime` and `runtime-async` packages, I did not implement proper handling for the case where `SAVEPER` is anything but 1.  (This is because both En-ROADS and C-ROADS currently use a `SAVEPER` of 1.)  There are other models out in the wild that use a non-1 `SAVEPER` so we should make sure the runtime packages handle this case properly.

PR #296, opened 2022-12-09
feat: expose accessor functions for control parameters

### LINKED ISSUE #295: Expose accessor functions for control parameters
To help enable the other changes and refactoring being done to support non-1 `SAVEPER` values in #291, I'm proposing to change the code gen to emit 3 accessors/getters for `INITIAL TIME`, `FINAL TIME`, and `SAVEPER`.  These values are needed by the `runtime` and `runtime-async` packages when allocating the buffers that will hold the output values.  The generated C code would look like this:

```
// Control parameter accessors
double getInitialTime() {
  return 2000;
}
double getFinalTime() {
  return 2100;
}
double getSaveper() {
  return 1;
}
```

Once these are available, we'll be able to simplify the `build` and `runtime` packages:
* Currently the `build` package expects the developer to provide the initial/final times in the `sde.config.js` file, but this is error prone (because they can get out of sync with the actual values in the mdl) and also just redundant configuration.  It shouldn't be necessary to pass these in as long as we can read them from the generated model.
* Once these accessors are available, the `runtime` package can read them after initializing the wasm module, and crucially, we'll have access to the `SAVEPER` value so that we can fix the few other places that are currently assuming it to be 1.

I have the code gen changes set up such that it expects `INITIAL TIME` and `FINAL TIME` to be constant values, and `SAVEPER` can either be a constant numeric value or equal to `TIME STEP` (which is also expected to be constant).  I think these are valid assumptions given the way the rest of the native code works.

chrispcampbell:
Fixes #295 

This changes the code gen to emit 3 new accessors that allow the `runtime` and `runtime-async` packages to more easily read the control parameters (`INITIAL TIME`, `FINAL TIME`, and `SAVEPER`) prior to running the model for correctly sizing the allocated buffers that hold output values.  See issue for more details.

There are no tests included here, but I've already verified that this works on my other branch for #291, and that branch also includes automated JS-level integration tests that verify the behavior.

@ToddFincannonEI: It would be great if you could review this change if you've got time.  I decided to split it out from my other changes for #291 (supporting non-1 `SAVEPER` values in the runtime packages) because these 2 files are the only ones that I touched in the `cli` and `compile` packages.  That way you only have to review 2 files, and I'll submit/merge the other build/runtime changes once this PR is merged.  Also, having these changes broken out makes it easier to highlight this as a separate "feature" in the changelog that's independent from the other changes.

### DIFF
--- a/packages/cli/src/c/sde.h
+++ b/packages/cli/src/c/sde.h
@@ -50,7 +50,7 @@ EXTERN double _final_time;
 EXTERN double _time_step;
 EXTERN double _saveper;
 
-// API
+// API (defined in model.h)
 char* run_model(const char* inputs);
 void runModelWithBuffers(double* inputs, double* outputs);
 void run(void);
@@ -58,7 +58,12 @@ void startOutput(void);
 void outputVar(double value);
 void finish(void);
 
-// Functions implemented by the model
+// API (defined by the generated model)
+double getInitialTime(void);
+double getFinalTime(void);
+double getSaveper(void);
+
+// Functions implemented by the generated model
 void initConstants(void);
 void initLevels(void);
 void setInputs(const char* inputData);
--- a/packages/compile/src/generate/code-gen.js
+++ b/packages/compile/src/generate/code-gen.js
@@ -42,6 +42,7 @@ let codeGenerator = (parseTree, opts) => {
       } else if (operation === 'generateC') {
         // Generate code for each variable in the proper order.
         let code = emitDeclCode()
+        code += emitGetControlValuesCode()
         code += emitInitLookupsCode()
         code += emitInitConstantsCode()
         code += emitInitLevelsCode()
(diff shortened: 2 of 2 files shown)


ToddFincannonEI:
This worked fine for me in some tests on a simple model. The correct values were written into the generated accessors.

I see that `TIME STEP` is not surfaced in an accessor, so I assume that is unnecessary for the `build` and `runtime` packages .

The assumptions about the possible values of `SAVEPER` seem correct to me.

ToddFincannonEI (approved):
This worked fine for me in some tests on a simple model. The correct values were written into the generated accessors.

I see that `TIME STEP` is not surfaced in an accessor, so I assume that is unnecessary for the `build` and `runtime` packages .

The assumptions about the possible values of `SAVEPER` seem correct to me.

chrispcampbell:
> I see that `TIME STEP` is not surfaced in an accessor, so I assume that is unnecessary for the `build` and `runtime` packages.

That's correct, I decided not to expose it because it's not needed by those packages, and I couldn't think of a use case where a developer would need access to it (since it mainly controls the internal behavior of the model).  We could always add it later if we find a need for it.

Issue #302, opened 2022-12-12
Revisit interface between JS runtime package and generated C model


chrispcampbell:
One of the things I'd like to address before we have a "1.0" of SDE is the exported API of the C model.  I'm not planning to make these changes right away, but wanted to create this issue to start documenting/discussing.

Originally there was one primary entrypoint in the C API that worked with string data (this one used `setInputs` in the generated model):
```c
char* run_model(const char* inputs);
```

Later I added an alternate entrypoint that takes pointers to pre-allocated `double` buffers to have more optimal communication between the JS-level runtime code and the wasm model (this one used a new `setInputsFromBuffers` in the generated model):
```c
void runModelWithBuffers(double* inputs, double* outputs);
```

Besides the awkward naming of the new ones (to avoid conflicting with the old), it would be preferable to have only one way of doing things, both to simplify the implementation and to make the API less confusing.

I think we could probably reimplement the old string-based API as a higher-level thing that's written to use a lower-level one (more like `runModelWithBuffers`).

There may be other changes we could make to the C API (in `model.c`) and in the generated C code that is related to the above; we can note them in comments in this issue and then maybe figure out a plan for this once we have time to think things through.

chrispcampbell:
As a secondary thing, we might want to (re)consider how the control parameters (`INITIAL TIME`, `FINAL TIME`, and `SAVEPER`) are initialized and made available to the higher-level code.  See related discussion in the two recent issues about the control parameter accessors (#295 and #301).

For example, instead of implicitly initializing these in the new accessors, it might make sense to (also) provide an explicit way of initializing them, like a `initControlParameters` function.  The benefit of this is that the higher level code could potentially override the values in the model, or set them based on inputs (similar to how `runModelWithBuffers` takes a list of input values).

PR #304, opened 2022-12-12
fix: initialize control parameters prior to access

### LINKED ISSUE #301: Control parameter accessors need to generate code
I was wrong when I said that the new control parameter accessor values would always be constants, like they are in most models. EPS is highly data-driven to adapt to different model regions. The initial and final times are read from CSV files. Simply copying the formula text doesn't work. We need to generate code for the RHS after all.

The attached zip file illustrates the problem with a model adapted from EPS. Here's how the initial and final times are specified.

[control.zip](https://github.com/climateinteractive/SDEverywhere/files/10199326/control.zip)

```
INITIAL TIME=
	GET DIRECT CONSTANTS(
	   'InputData/IT.csv',
	   ',',
	   'B2*'
	)
	~	Year
	~	The initial time for the simulation.
	|

FINAL TIME=
	GET DIRECT CONSTANTS(
	   'InputData/FT.csv',
	   ',',
	   'B2*'
	)
	~	Year
	~	The final time for the simulation.
	|
```

This results in the following invalid C code for the new control parameter accessors.

```c
double getInitialTime() {
  return GET DIRECT CONSTANTS('InputData/IT.csv',',','B2*');
}
double getFinalTime() {
  return GET DIRECT CONSTANTS('InputData/FT.csv',',','B2*');
}
```

Is it possible to define the accessors like this? Or are they not initialized yet when the runtime needs them?

```c
double getInitialTime() {
  return _initial_time;
}
double getFinalTime() {
  return _final_time;
}
double getSaveper() {
  return _saveper;
}
```

chrispcampbell:
Fixes #301 

With this change we no longer generate the control param accessors as part of code gen, but instead define accessors in `model.c` that return the computed values (e.g., `_initial_time`).  For this to work correctly in cases where the model's control parameters are computed (or initialized based on external data files), we run the model for the first time step only, and we do this lazily on first use of the accessors.

No change in behavior for models that use constant values for the control params.  This fix mainly applies to models that define the control params in an external file, like EPS.

I added a JS-level integration test that verifies the behavior for the latter case where they are defined in an external csv file.  For this test, we actually cover a mix of external and computed params:
* `INITIAL TIME` and `TIME STEP` are defined in a csv file
* `FINAL TIME` is computed based on `INITIAL TIME`
* `SAVEPER` is computed based on `TIME_STEP`

### DIFF
--- a/packages/cli/src/c/model.c
+++ b/packages/cli/src/c/model.c
@@ -23,6 +23,49 @@ size_t savePointIndex = 0;
 
 int step = 0;
 
+void initControlParamsIfNeeded() {
+  static bool controlParamsInitialized = false;
+  if (controlParamsInitialized) {
+    return;
+  }
+
+  // Some models may define the control parameters as variables that are
+  // dependent on other values that are only known at runtime (after running
+  // the initializers and/or one step of the model), so we need to perform
+  // those steps once before the parameters are accessed
+  // TODO: This approach doesn't work if one or more control parameters are
+  // defined in terms of some value that is provided at runtime as an input
+  initConstants();
+  initLevels();
+  _time = _initial_time;
+  evalAux();
+  controlParamsInitialized = true;
+}
+
+/**
+ * Return the constant or computed value of `INITIAL TIME`.
+ */
+double getInitialTime() {
+  initControlParamsIfNeeded();
+  return _initial_time;
+}
+
+/**
+ * Return the constant or computed value of `FINAL TIME`.
+ */
+double getFinalTime() {
+  initControlParamsIfNeeded();
+  return _final_time;
+}
+
+/**
+ * Return the constant or computed value of `SAVEPER`.
+ */
+double getSaveper() {
+  initControlParamsIfNeeded();
+  return _saveper;
+}
+
 char* run_model(const char* inputs) {
   // run_model does everything necessary to run the model with the given inputs.
   // It may be called multiple times. Call finish() after all runs are complete.
--- a/packages/cli/src/c/sde.h
+++ b/packages/cli/src/c/sde.h
@@ -50,18 +50,16 @@ EXTERN double _final_time;
 EXTERN double _time_step;
 EXTERN double _saveper;
 
-// API (defined in model.h)
+// API (defined in model.c)
+double getInitialTime(void);
+double getFinalTime(void);
+double getSaveper(void);
 char* run_model(const char* inputs);
 void runModelWithBuffers(double* inputs, double* outputs);
 void run(void);
 void startOutput(void);
 void outputVar(double value);
 void finish(void);
-
-// API (defined by the generated model)
-double getInitialTime(void);
-double getFinalTime(void);
-double getSaveper(void);
 
 // Functions implemented by the generated model
 void initConstants(void);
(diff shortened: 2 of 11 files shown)


chrispcampbell:
@ToddFincannonEI: If you can review the cli/compile changes, that would be great.  The new integration test is mostly copy/pasted from the previous `saveper` test, so you don't need to worry as much about those changes.  (I'd like to simplify the JS-level integration tests eventually to not require so much boilerplate, but I'll worry about that later once we have more of them.)

ToddFincannonEI (commented):
I'm wondering if `initConstants`, `initLevels`, and `evalAux` would be called twice — here in the first call to `initControlParameters`, and then in `runModel`. I am pretty sure those calls are idempotent, but it could amount to a lot of run time in a large model. I might be missing how this works.

I also don't set `setInputs` getting called in `initControlParameters`, so I don't understand how inputs could affect the control parameters.

chrispcampbell:
> I'm wondering if `initConstants`, `initLevels`, and `evalAux` would be called twice — here in the first call to `initControlParameters`, and then in `runModel`. I am pretty sure those calls are idempotent, but it could amount to a lot of run time in a large model. I might be missing how this works.

Yes, those will run once (and only once) as part of `initControlParameters`, so the first time they're called in `run` will be redundant.

My assumption is that even for a large model, the cost of running these functions once at init time is relatively tiny.  Using En-ROADS as an example, a single model run includes ~880 time steps and runs in about 20 ms, so this init code might take around 0.02 ms.  There are probably other larger models where it could matter more, but my thinking is that this fix is more of a stopgap just to get us past the current regression affecting EPS, and we can improve upon it later.

I considered other alternative approaches that would avoid the redundancy.  For example:
* We could make code gen produce a separate init function that includes the subset of `initConstants` and `evalAux` that are needed to initialize the control params without dragging in all the rest.
* We could change `run` to skip the first `initConstants` if we know that it was already done as part of the `initControlParams` (but that won't save much time).
* We could go back to making the user/developer have to provide values at config/init time (error prone and not very friendly).

But I think all of those are more involved than the simplest fix we can do for now.

> I also don't set `setInputs` getting called in `initControlParameters`, so I don't understand how inputs could affect the control parameters.

That TODO comment I included is referring to the fact that we don't factor in runtime inputs there (like we do for the normal `run` function).  Maybe I could've worded the comment more clearly.  I think the more important thing (since TODO comments are overlooked) is that I added this deficiency as a [comment](https://github.com/climateinteractive/SDEverywhere/issues/302#issuecomment-1347391672) in #302 to be considered later.

To put it another way, I think we agreed to punt for now on handling the possibility of a model whose control parameters are influenced by inputs provided at runtime, correct?

ToddFincannonEI:
Thanks for the explanation. I thought you had refactored to support control parameter inputs, but I see now that has been deferred. The extra init calls are a bigger hit for EPS, which usually only has 30 time steps, but initializes a huge number of variables generated from data files. However, I don't think it will affect EPS, because we are not using the SDEverywhere runtime yet. So I'm fine with completing this change as written.

chrispcampbell:
OK, thanks for the review.  I'll merge these changes shortly to resolve the regression, but I've filed a separate issue #305 to revisit this fix and I'll try to come up with a less heavy-handed approach to initializing these parameters.

PR #309, opened 2022-12-16
fix: read blank lines and blank cells in CSV files

### LINKED ISSUE #308: CSV files with blank lines read incorrectly
`readCSV` skips rows in CSV files that have all empty fields. Note that Excel writes such rows with comma delimiters but no data between delimiters, like this: `,,,`. This throws off cell references in Vensim functions such as `GET DIRECT CONSTANTS`. 

The attached model illustrates the problem.

[csv-blank-line.zip](https://github.com/climateinteractive/SDEverywhere/files/10249993/csv-blank-line.zip)

For this model, `sde test` prints:
```
_k time=0.00 vensim=3 sde=4 diff=33.333333%
```

ToddFincannonEI:
Fixes #308 

I enabled 'csv-parse` flags to not skip empty cells and empty rows in `readCsv`. This fixes the test case in the issue. EPS is building OK with this change.

I don't know why these flags were disabled in the first place. Vensim formulas contain absolute cell references that will be incorrect if empty rows and cells are skipped.

The `test:c-int` model tests run OK. I wanted to add the test in the issue to `directconst.mdl`, but the recent `h.csv` file with non-numeric cell gives a fatal error in Vensim 9.2.1, so I could not produce the reference `.dat` file.


chrispcampbell:
@ToddFincannonEI:

> I wanted to add the test in the issue to `directconst.mdl`, but the recent `h.csv` file with non-numeric cell gives a fatal error in Vensim 9.2.1, so I could not produce the reference `.dat` file.

I had left a comment in `directconst.mdl` regarding how to deal with the "invalid cell" test:

```
h[DimB, DimC] =
  GET DIRECT CONSTANTS(
    'data/h.csv',
    ',',
    'B2'
  )
  ~
  ~ This csv file has a cell that does not contain a number so that we can check that
    SDE does not throw an error in this case.  Note that Vensim will raise an error if
    a cell in the csv file does not contain a number, so to run this test in Vensim,
    temporarily change cell C3 to 0 in `h.csv`.
```

But I recognize that it was subtle and easy to overlook, so I moved that to its own model test (`directconst_invalidcell`) to avoid polluting the normal `directconst` test.  After doing that, I took your test that was attached to the issue and folded it into the existing `directconst` test.  So now we have test coverage for both issues.

The fix itself looks fine, so I'll merge this all shortly.

chrispcampbell (approved):
Looks good.

PR #317, opened 2023-04-25
fix: code generation with spec and model using GET DIRECT DATA and/or GET DIRECT LOOKUPS

efad2022:
Issue: if a model has direct lookups, like the following:
`sample value = GET DIRECT LOOKUPS( 'sample.csv' , ',' , 'A', 'C2' )`
then when generating the code with a spec the variable is removed (no code generated).

### DIFF
--- a/packages/compile/src/model/model.js
+++ b/packages/compile/src/model/model.js
@@ -361,8 +361,9 @@ function removeUnusedVariables(spec) {
     }
   }
 
-  // Filter out unneeded variables so we're left with the minimal set of variables to emit
-  variables = R.filter(v => referencedVarNames.includes(v.varName), variables)
+  // Keep vars with directDatArgs (GET DIRECT DATA and GET DIRECT LOOKUPS) 
+  // and filter out unneeded variables so we're left with the minimal set of variables to emit
+  variables = R.filter(v => !R.isEmpty(v.directDataArgs) || referencedVarNames.includes(v.varName), variables)
 
   // Rebuild the variables-by-name map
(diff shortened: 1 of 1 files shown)


chrispcampbell:
@efad2022: I was just revisiting this PR and have a question.  Do you have a small Vensim model and/or `spec.json` file that I can use to understand the issue?

One thing to note about using a `spec.json` file (specifically with `outputVarNames`) is that `sde` will keep equations in the generated C file only if they are referenced directly or indirectly by `outputVarNames`.

So for example, if you have a model like this:
```
sample value = GET DIRECT LOOKUPS( 'sample.csv' , ',' , 'A', 'C2' ) ~~|
x = sample value ~~|
y = 10 ~~|
```

If your `spec.json` has `outputVarNames: ["sample value"]`, then the lookup variable should be retained in the generated C code because it is referenced directly.

If your `spec.json` has `outputVarNames: ["x"]`, then the lookup variable should also be retained in the generated C code because it is used by `x`.

If your `spec.json` has `outputVarNames: ["y"]`, then the lookup variable will not be retained in the generated C code because neither `x` nor `sample value` are referenced.

Let me know if this makes sense or if I'm overlooking something.

chrispcampbell:
It's not clear if there was an actual bug here (see my [previous comment](https://github.com/climateinteractive/SDEverywhere/pull/317#issuecomment-1815620099)) and I haven't received new details so I think it's best to close this.  If more details are provided to reproduce an issue, we can file an issue and prepare a new fix.

PR #333, opened 2023-06-20
docs: reorganize README with emphasis on end user documentation

### LINKED ISSUE #265: Reorganize README
The current README is very thorough, but it contains a mix of user-facing (installation, usage) and developer-facing (debugging, architectural) documentation, so it may be overwhelming to new users.  Both kinds of documentation are important, but I think it would be better to have the succinct user-facing docs in the README, and move the developer-facing docs to the wiki or a separate `docs` folder.

Additionally, all the new packages that have been added in recent months haven't yet been mentioned in the README, mostly because I've been waiting to revisit how the README is organized.

For the new README, I propose the following structure:
* Intro
* Caveats
* Quick Start (`npm create @sdeverywhere`)
* Packages (table of `@sdeverywhere` packages, with links to docs+changelog for each)
* Installation and Usage
    * Using SDE to build a web app (high-level)
    * Using SDE to generate C code (low-level)
* Supported Functions (list of Vensim functions supported by SDE, with caveats noted)

As part of this, I will need to revisit the "Using SDEverywhere to Make a Vensim Model into a Web Application" document and update it to account for the new build/plugin system, or just fold the changes into the README.

Checklist of things to resolve as part of this work:
- [x] Revisit "Using SDEverywhere to Make a Vensim Model into a Web Application" document
- [x] Include note about `DEPRECIATE STRAIGHTLINE` fiscal period argument being unsupported (see https://github.com/climateinteractive/SDEverywhere/pull/264#pullrequestreview-1153710730)
- [x] Change Quick Start instructions to suggest cd to a directory containing an mdl file

chrispcampbell:
Fixes #265 

Summary of changes:
* Rewrote the top-level README to cover newer functionality that has been added in the past couple years (model check tools, local dev environment, the `create` script, etc).
* Added screenshots and videos.
* Moved the more technical developer/contributor-focused documentation from the top-level README to a new section for contributors in the wiki.
* Moved the detailed `sde` command reference to the `cli` package README.
* Removed the documentation files that were previously in the `notes` directory; these have been migrated to the wiki.  (I haven't finished rewriting the "making a web app" guide, but there's a placeholder on the wiki that should be OK for now and I'll finish fleshing that page out after this gets merged.)
* Added consistent sections (Quick Start, Install, Usage, Documentation) to the READMEs for individual packages.

See the rendered README for this branch here (i.e., what the landing page will look like after this branch is merged):
https://github.com/climateinteractive/SDEverywhere/blob/chris/265-readme/README.md


ToddFincannonEI (commented):
I love the new introduction. It makes me excited about using SDEverywhere. Getting something to run with one command is a much lower barrier to entry that what we had before. I like the way you separated out the more technical details into the wiki.

Here are a few questions and comments by section.
### Supported Platforms
- Have we run WebAssembly on the server?
### Supported File Formats
- Is there a published roadmap? We may not want to mention XMILE given the low priority in both our ourgs.
- This would be a good place to link to the supported Vensim functions in the wiki.
- It would be helpful to document someplace (probably not here) the compiler error you get when using an unsupported Vensim function. This seems to surprise people, and it looks like SDEverywhere generated bad code. Maybe just issue an SDEverywhere compiler message instead.
### Features
- Live editing and quality checks and comparisons might be our killer feature, since your tools do this better than the vendor tools. Maybe mention it up higher.
### SDEverywhere in Production
- You could add EPS to the production apps. [Energy Policy Simulator](https://energypolicy.solutions/)
### Contributing
- We fully handle :EXCEPT: clauses now.
- Should we mention antlr4-vensim?

ToddFincannonEI (approved):

chrispcampbell:
@ToddFincannonEI: Thanks for the review and ideas.  I pushed a bunch of commits to address your feedback, and comments are inline below.  I'll merge this PR shortly and we can continue to improve the README as things come up.

> I love the new introduction. It makes me excited about using SDEverywhere. Getting something to run with one command is a much lower barrier to entry that what we had before. I like the way you separated out the more technical details into the wiki.

Great!

> ### Supported Platforms
> 
>  * Have we run WebAssembly on the server?

Yes, the same WebAssembly blob can run in both browser and in a command-line or server-side Node application.  There are examples that exercise this in a few places, for example, in the integration tests (under `tests/integration`, like [this script](https://github.com/climateinteractive/SDEverywhere/blob/main/tests/integration/saveper/run-tests.js)).  When model-check tests are run on the command line, that's another example of the WASM code running in a Node process.

Also, the En-ROADS SDK includes a simple example of running the model in a Node application.  Your question makes me realize that we can include some better docs and examples for this use case; we can even make the `create` script auto-generate a small Node script that runs the model in addition to the web app we already generate.  I'll add it to my list.

> ### Supported File Formats
> 
> * Is there a published roadmap? We may not want to mention XMILE given the low priority in both our ourgs.

The only roadmap at this point is [this post](https://github.com/climateinteractive/SDEverywhere/discussions/199) I wrote up on the discussion board, along with the roadmap that's in my head.  I was on the fence about mentioning XMILE for the reason you mention, but thought it might be worth mentioning in case anyone is inspired to help make it a reality.

I changed it to not say "roadmap"; instead it now says that it's not currently supported but is something we'd like to see if anyone wants to help.

> * This would be a good place to link to the supported Vensim functions in the wiki.

Good idea.  Link added.

> * It would be helpful to document someplace (probably not here) the compiler error you get when using an unsupported Vensim function. This seems to surprise people, and it looks like SDEverywhere generated bad code. Maybe just issue an SDEverywhere compiler message instead.

Yeah, in the short term we can document this, but I think ideally it would be better if the SDE compiler would fail fast and emit a more useful error message in this case.  I'll file an issue this week.

> ### Features
> 
> * Live editing and quality checks and comparisons might be our killer feature, since your tools do this better than the vendor tools. Maybe mention it up higher.

Yeah, I wasn't sure the best place for the Features section (since it's a little wordy), but having it up higher makes sense.  I've moved it up to be between "Supported File Formats" and "Core Functionality".

> ### SDEverywhere in Production
> 
> * You could add EPS to the production apps. [Energy Policy Simulator](https://energypolicy.solutions/)

Great, I meant to ask you directly if it would be OK to include EPS.  I've added a bullet.  Let me know if you'd like to change the wording.

> ### Contributing
> 
> * We fully handle :EXCEPT: clauses now.

Good catch, I removed mention of EXCEPT clauses.

> * Should we mention antlr4-vensim?

Sounds good, I added a link to the first bullet about improving the parser.

PR #350, opened 2023-09-02
fix: change ramda imports to wildcard style

### LINKED ISSUE #349: Update ramda imports
I was working on adding some new tests in the `compile` package (to expand coverage of existing behavior) and found that vitest doesn't work with the existing import style for ramda.

We currently use:
```js
import R from 'ramda'
```

We are currently on ramda@0.27.2 and have been for some time, but I found in the 0.25 upgrade guide that when they changed to ESM, they said the import style needs to be changed to:
```js
import * as R from 'ramda'
```

See 0.25 upgrade guide:
https://github.com/ramda/ramda/issues/2319

Once I made this change, then vitest has no problem importing all the `compile` package sources that use ramda.  And it has no impact on behavior (tests all pass).

chrispcampbell:
Fixes #349 

@ToddFincannonEI: See issue for more details.  No impact on `sde` behavior but needed to unblock testing via vitest.

### DIFF
--- a/packages/cli/src/sde-compare.js
+++ b/packages/cli/src/sde-compare.js
@@ -1,7 +1,7 @@
 import { existsSync } from 'fs'
 
 import { pr } from 'bufx'
-import R from 'ramda'
+import * as R from 'ramda'
 
 import { readDat } from '@sdeverywhere/compile'
 
--- a/packages/cli/src/sde-log.js
+++ b/packages/cli/src/sde-log.js
@@ -3,7 +3,7 @@ import path from 'path'
 
 import B from 'bufx'
 import byline from 'byline'
-import R from 'ramda'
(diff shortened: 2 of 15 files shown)


ToddFincannonEI (approved):
I tested the new import syntax separately and found it worked in the latest Ramda but not 0.24.1. We are using Ramda 0.27.2 and could consider bumping to 0.29.0, but that does not look necessary at the moment.

chrispcampbell:
> I tested the new import syntax separately and found it worked in the latest Ramda but not 0.24.1.

Correct, it would only work in 0.25 after they migrated to ESM.

> We are using Ramda 0.27.2 and could consider bumping to 0.29.0, but that does not look necessary at the moment.

Yep, we can save that for another day; if it ain't broke... :)

Issue #319, opened 2023-04-29
Add WebGPU Support


jrissman:
WebGPU is a new API that lets web applications leverage the power of modern hardware for general purpose computation.  After six years of development, it was just released in beta on Chrome for Windows, Mac, and ChromeOS.  All major browsers are currently working on adding support (it can already be enabled in Safari, Firefox, Edge, etc. with a flag), and Android and iOS support are coming as well.  For more, see the beta release announcement here: https://developer.chrome.com/blog/webgpu-release/

WebGPU might enable much faster model run speeds on modern devices.  This might allow models to perform more calculations (such as full 8760-hour electricity dispatch) while still running the model quickly enough to feel interactive.  It would also be useful for optimization and genetic algorithms, where a model may need to be run many times, with the computer varying the settings each time, to home in on an optimal solution defined by a goal function.

A modern GPU can have over 1,000 cores, vastly more than a CPU.  Assuming WebGPU abstracts the process of distributing work among GPU cores, runspeed might increase by orders of magnitude.

WebGPU can be invoked by WebAssembly code generated by Emscripten.  Here's a repo I found where someone built tools to do this: https://github.com/juj/wasm_webgpu .  Therefore, it appears that WebGPU is not a replacement for WebAssembly but rather could be added on for further speed gains.

It would be useful to investigate how easy or difficult it would be to use the WebGPU API and see what sort of runspeed improvement it provides.  If the runspeed improvement is dramatic, it might be worthwhile to plan to migrate to WebGPU as the standard method of running models once the WebGPU API is broadly supported across browsers and operating systems, as it likely will be in a couple years.

ToddFincannonEI:
This is something that could have a big payoff for large models. Developing a demo with Emscripten could be a quick project. I think it would need to be optional for the code generator until it matures. Jeff's proposal contains two use cases — interactive speedup on a personal computer, which would have a modest number of GPU cores; and a big speedup when running models on a cloud server with ~1000 GPU cores. I don't think anyone has used SDEverywhere the second way, but it is certainly a possibility for either C or WebAssembly code.

jrissman:
Another way to look at it is in floating point operations per second, where GPUs have orders-of-magnitude edge over CPUs.  Here's a chart from Cornell University:

![PeakFlopsCPUGPU](https://user-images.githubusercontent.com/7120106/235499517-31c7b8b3-ed2a-4fdf-8517-a32ea9714db1.png)

[...]

PR #355, opened 2023-09-09
feat: add support for capturing data for any variable at runtime

### LINKED ISSUE #105: Add support for capturing data for any variable at runtime
When a `spec.json` file is used to expose a subset of output variables, only those specified variables can be accessed as outputs.  For a testing tool we are working on internally, we would like to be able to inspect the data for _any_ variable in the model, not just the ones exposed as outputs.  The En-ROADS model uses thousands of variables; if we listed them all as outputs, it would use a lot of memory and would slow down the model.  It would be preferable to have a way at runtime to select the variables for which to capture data into the output buffer.

I'm envisioning a few parts to this:

1. Add a new `sde generate --listjson` option that outputs a JSON file containing information about all variables and subscripts used in the model.  (This would also be useful for a dependency tree tracking tool that we developed internally.)

2. Accept a new `outputIndices` buffer parameter:
```
void runModelWithBuffers(double* inputs, double* outputs, int32_t* outputIndices) {
```

3. Generate a new `storeOutput` function that includes a big switch statement, which will select the output (and subscripts) by index based on the values from the `outputIndices` buffer:
```
void storeOutput(size_t varIndex, size_t subIndex0, size_t subIndex1, size_t subIndex2) {
#define USE_OUTPUT_INDICES 0
#if USE_OUTPUT_INDICES
  switch (varIndex) {
    case 1:
      outputVar(_a);
      break;
    case 2:
      outputVar(_b);
      break;
    case 3:
      outputVar(_c[subIndex0][subIndex1]);
      break;
    // ...
    default:
      break;
  }
#end
}
```

The flag would be set to 0 by default so that there would be no impact on size/speed unless you opt-in to use this new output mode.

Things to revisit before finalizing this work:
- [ ] Allow `--genc` and `--list` in the same command so that we don't have to run `sde` twice
- [ ] Exclude variables from `storeOutput` if they have more than 3 dimensions (we only support variables with <= 3 dimensions for now)

chrispcampbell:
Fixes #105 

@ToddFincannonEI: I'm finally getting around to submitting a PR for this feature that I started over 2 years ago.  We've been using this branch in production (for En-ROADS and C-ROADS) for the past 6 months, so I think it's stable enough to merge into `main`.

There are currently two use cases at CI where this feature comes in handy:
* In our automated model checks, we use this to access internal variables that aren't normally accessible to model-check (things like internal/historical data that aren't declared as output variables).
* For the new online [En-ROADS Technical Reference](https://docs.climateinteractive.org/projects/en-roads-reference-guide/en/latest/index.html), we use this to generate the hyperlinked [Model Equations](https://docs.climateinteractive.org/projects/en-roads-reference-guide/en/latest/pages/demand.html#demand__model_equations) component in each chapter.  This component allows anyone to see a graph of any equation in the model.

I realize this is a large PR, and there's no rush on reviewing it.  We could wait to go through it together when we meet in person soon.  I probably could've split the JSON listing part into its own PR, but it all kind of goes together so decided to leave it in.

I'll try to succinctly explain how it all fits together:

1. `sde generate --list` now produces a JSON file in addition to the existing YAML and TXT outputs.  I know we had talked about removing the YAML and TXT variants but I figure it's better to do that as a separate change since this one is already big enough.

1. In that JSON file, for each variable, there is a `varIndex` field.  This `varIndex` corresponds to the value that code gen emits in the new `storeOutput` function in the generated C code.  For example, variable `X` might have a `varIndex` of 1.  At runtime, the code that runs the model can request data for variable `X` by using that index value of 1.

1. The code gen outputs a giant switch statement in the generated C code that handles the mapping of `varIndex` values to the appropriate internal C variable (and handles access to subscripted variables as well).  The `storeOutput` function body is gated by a preprocessor flag that is off by default (it needs to be enabled at build time).  Therefore, even though the generated C code is larger, there is no impact on the compiled code size or performance when the switch is off.

1. Likewise, there's a new `outputIndexBuffer` argument to the `runModelWithBuffers` function, but in the common case where this parameter is left as `NULL`, then that code is completely unused and there is no impact on runtime performance or behavior.

1. The JS-level `runtime` and `runtime-async` packages have been updated to support this new functionality, but I decided to apply `@hidden` tags for most of the new APIs so that they don't yet appear in the public API documentation.  This is because a) I'm not ready to call this new API "done", as I've got some changes in mind to make it better and b) we (CI) are probably going to be the only users of this API in the beginning.  Once this initial work gets merged in, I can look at the API again with fresh eyes (and ideally align it with how we incorporate your proposal for changes to how inputs are passed in).

1. I added an automated E2E test of the new functionality in `tests/integration/impl-var-access` to make sure it all works together.


travisfranck:
Cool new change.  I can definitely see how this will be useful for unit tests.

ToddFincannonEI:
I haven't reviewed the code, but I will comment on your description. This is not something I would use in my own work, but it's a big step toward duplicating the full functionality of Vensim. Having the giant switch statement off by default is important for me, since our models are so large. 

I would like to keep the text listing output. That's something I use a lot. I think the JSON output could replace YAML. I have some model analysis tools that use YAML, but they could use JSON instead. I agree that JSON is a better choice than YAML for model listings.

chrispcampbell:
@ToddFincannonEI:

> I haven't reviewed the code

Not a problem.  Let me know if you mean "haven't reviewed yet" (in which case I'll await your review) or if you mean "unlikely to review".

> Having the giant switch statement off by default is important for me, since our models are so large.

Yes, same here.

FWIW, I think I could get rid of the "giant switch statement" approach entirely if the code generator had support for pluggable/custom allocators (which I alluded to in a [comment](https://github.com/climateinteractive/SDEverywhere/issues/319#issuecomment-1530117453) on the WebGPU feature).

But that's a separate topic.  For now, with the switch off, there's no impact on build or runtime performance.

> I would like to keep the text listing output. That's something I use a lot. I think the JSON output could replace YAML.

Fair enough, all three are still available.  Later (separate from this) we can drop the YAML output in favor of JSON, and we can keep the TXT output.

ToddFincannonEI:
Sounds good — I am unlikely to review the code itself.

PR #375, opened 2023-10-16
fix: allow numeric ranges in subscript range definitions

### LINKED ISSUE #318: Subscript numeric range fails inside an index list
Subscript ranges can include a [numeric range](https://www.vensim.com/documentation/22090.html) that abbreviates a sequence of subscript indices as in the ref sample model.

```
Target: (t1-t3)
```

The code generator only handles this when the numeric range is the entire subscript definition as shown above. It fails when the numeric range occurs inside a subscript index list like this:

```
DimC: C1, (C2-C4), C5
t[DimC] = 1
```

Fixing this will require an [antlr4-vensim](https://github.com/climateinteractive/antlr4-vensim) grammar change and corresponding support in the code generator.

ToddFincannonEI:
Fixes #318 

This PR works with an update to the antlr4-vensim parser (todd/9-numeric-range) to support numeric ranges in subscript range definitions. See SDE issue #318 and https://github.com/climateinteractive/antlr4-vensim/issues/9 for details. I added a test in `subscript.mdl` to illustrate the issue.


ToddFincannonEI:
I refactored `visitSubscriptList` to match antlr4-vensim branch `todd/9-numeric-range`. I tested this fix with EPS code generation.

chrispcampbell (approved):
Looks good, I'll merge it shortly.

PR #379, opened 2023-10-17
fix: correct handling of lookups defined by GET DIRECT LOOKUPS when used as function call arguments

### LINKED ISSUE #378: Incorrect code gen using GET DIRECT LOOKUPS lookups
The `LOOKUP FORWARD`, `LOOKUP BACKWARD`, and `LOOKUP INVERT` functions takes a lookup as its first argument. This works fine when the lookup is defined using the literal lookup syntax, as in our existing test. However, when the lookup is read from a file with `GET DIRECT LOOKUPS`, incorrect code is generated: the lookup is dereferenced with `_LOOKUP` instead of being passed to the function.

ToddFincannonEI:
Fixes #378 

See issue for details.


chrispcampbell (approved):
Looks good.  I have a branch that's nearly ready with a ton of unit tests for equation reading and code gen, so I will make sure this is also covered in those tests.

PR #394, opened 2023-11-14
build: only run build workflow for push events OR pull_request events from fork

chrispcampbell:
When Travis submitted a PR a couple weeks ago from an external fork, I added a temporary fix to the `build.yaml` workflow so that it runs for both `push` and `pull_request` events, but this leads to redundant builds in the more typical case where the workflow is run for a branch of the primary SDEverywhere repo.

I'm attempting to address this by skipping jobs in the case where it's a `pull_request` event for a branch of this repo.  We'll see if it works after I open this PR.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -25,6 +25,14 @@ env:
 jobs:
   # The `build` job builds and tests all packages in the monorepo
   build:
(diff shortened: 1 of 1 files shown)


chrispcampbell:
This results in some "skipped" jobs being reported in the PR checks section, but seems like a small price to pay.  I will merge this and we'll see if it can be improved further later.

image

PR #396, opened 2023-11-15
fix: allow lowercase cell reference in GET DIRECT function calls

### LINKED ISSUE #395: Allow lowercase cell reference in `GET DIRECT ...` function calls
I received a Vensim model that makes use of `GET DIRECT DATA` and `GET DIRECT CONSTANTS`, and the `cell` argument uses lowercase letters, like this:

```
	GET DIRECT CONSTANTS('?data', 'sheet' , 'c5' )
```

This works in Vensim, but does not currently work in SDEverywhere because the underlying library that reads xlsx files is not lenient.  If we convert the cell reference to uppercase, then it works fine.

As part of this, we should also handle the case where `XLSX.utils.decode_cell` fails to decode and throw a more useful error message.  In the current implementation, processing continues and you won't know about the failure until the C or WASM compiler reports a compilation error.

chrispcampbell:
Fixes #395 

This is a simple fix that allows `GET DIRECT ...` calls to use a cell reference that has lowercase letters instead of the more typical uppercase ones.  Vensim is lenient and allows either form, but the XLSX library we use in SDE is not lenient, so I've added a `toUpperCase()` call in the appropriate places.

I also added some error handling so that users will get a more useful error message in the unlikely case where an invalid cell reference is used (instead of allowing things to pass through).

I updated the existing integration tests and also added some new unit tests to verify this.

### DIFF
--- a/models/directconst/directconst.mdl
+++ b/models/directconst/directconst.mdl
@@ -18,8 +18,12 @@ b[DimB] =
   GET DIRECT CONSTANTS(
     'data/b.csv',
     ',',
-    'B2*'
-  ) ~~~:SUPPLEMENTARY|
+    'b2*'
+  )
+  ~
+  ~ This one uses a lowercase letter in the cell reference (atypical, but valid in Vensim).
+  ~ :SUPPLEMENTARY
+  |
 
 c[DimB, DimC] =
   GET DIRECT CONSTANTS(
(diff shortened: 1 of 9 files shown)


ToddFincannonEI (commented):
Looks like a straightforward fix — also good to see the extra error checks and new test coverage.

PR #398, opened 2023-11-15
fix: handle case where subscript/index appears in expression position

### LINKED ISSUE #397: Handle case where index (subscript) appears in expression position
I received a Vensim model where the modeler has an equation like this (simplified):

```
    DimA: A1, A2 ~~|
    x[DimA] = IF THEN ELSE(DimA = A2, 1, 0) ~~|
```

This works fine in Vensim, but fails in SDEverywhere.  We do handle the case where a dimension is used in expression position, but we don't currently handle the case where a subscript/index is used in expression position.

Note that this case is more commonly handled using separate equations (in my experience) like in the following sample, but there are probably valid use cases for subscripts in expression position, so we should fix SDE to allow the form above.

```
    DimA: A1, A2 ~~|
    x[A1] = 0 ~~|
    x[A2] = 1 ~~|
```

Or:

```
    DimA: A1, A2, A3 ~~|
    x[DimA] :EXCEPT: [A2] = 0 ~~|
    x[A2] = 1 ~~|
```

chrispcampbell:
Fixes #397 

See issue for more details.  I updated the existing "subscript" integration test and also added new unit tests so there is pretty good coverage for this fix.


ToddFincannonEI (commented):
I would probably have told them to rewrite their equation in idiomatic Vensim, but I agree it's good to have this capability for other cases, since it was easy to fix.

PR #400, opened 2023-11-15
fix: handle XLSX files in GET DIRECT {CONSTANTS,DATA,LOOKUPS}

### LINKED ISSUE #399: Handle XLSX files in `GET DIRECT {CONSTANTS,DATA,LOOKUPS}`
Currently SDEverywhere has incomplete support for XLSX files in the `GET DIRECT ...` functions.  See the "matrix" below.

We should unify the handling for these two functions so that any of the options supported by Vensim are also supported by SDEverywhere.

---

`GET DIRECT {DATA,LOOKUPS}` currently supports:
* an Excel file referenced using an indirect tag (`?data`), which must be set up in the `spec.json`, OR
* a CSV file referenced by file name (`'data.csv'`)

`GET DIRECT {DATA,LOOKUPS}` does not currently support:
* an Excel file referenced by file name (`'data.xlsx'`)

---

`GET DIRECT CONSTANTS` currently supports:
* a CSV file referenced by file name (`'data.csv'`)

`GET DIRECT CONSTANTS` does not currently support:
* an Excel file referenced using an indirect tag (`?data`), which must be set up in the `spec.json`, OR
* an Excel file referenced by file name (`'data.xlsx'`)

chrispcampbell:
Fixes #399 

See issue for more details.

I basically consolidated the handling of the 3 different kinds of tabular file sources:
* named csv file
* named xls[x] file
* tagged xls[x] file

And now any of those 3 sources work with any of:
* `GET DIRECT CONSTANTS`
* `GET DIRECT DATA`
* `GET DIRECT LOOKUPS`

Plenty of test coverage on this one.


ToddFincannonEI (commented):
I'm glad to see this cleaned up. The only potential problem I saw was in the  handleExcelOrCsvFile function, where the file extension is checked against lower case "csv". This would fail if the file name was upper case. I checked path.resolve and confirmed it does not normalize case.

chrispcampbell:
> I'm glad to see this cleaned up. The only potential problem I saw was in the handleExcelOrCsvFile function, where the file extension is checked against lower case "csv". This would fail if the file name was upper case. I checked path.resolve and confirmed it does not normalize case.

Good catch.  I updated the fix so that it converts the path to lowercase before checking the extension.  Also renamed one of the test files to have an uppercase CSV extension to verify this.  Will merge shortly.

PR #413, opened 2023-12-05
fix: add new implementation of read and code gen phases based on `parse` package

### LINKED ISSUE #412: Add new implementation of read and code gen phases based on `parse` package
I recently introduced a new `@sdeverywhere/parse` package in #405 that adds a new format-agnostic AST structure and a new parsing layer that insulates the `@sdeverywhere/compile` package from `antlr4-vensim`.

The next step (step 2 of the transition described in that issue) is to add new alternate implementations of the read and code gen phases of the compiler using the new `parse` package instead of depending directly on `antlr4-vensim`.

This work is now mostly complete (I've been working on it in a side branch for a while).  The goal was for the new implementations to be as close to 100% compatible (in terms of behavior and generated code) with the old implementations, and that goal has mostly been met (there are only a few slight differences in whitespace and use of parentheses).

My plan is to introduce the new code alongside the old code, with the new code disabled for now, but allow it to be enabled by setting a temporary environment variable (`SDE_NONPUBLIC_USE_NEW_PARSE=1`).

Once we are satisfied with the new code, I will file a third issue/PR to make the new code the default and remove the old implementation.

Here's the mapping of old source files to new ones in the `compile` package:

---

Preprocess phase + shared parse reader code:
* old:
    * preprocessor.js
    * model-reader.js
    * parser.js
* new:
    * (handled by `parse` package)

---

Read dimensions (aka subscript ranges) phase:
* old:
   * subscript-range-reader.js
* new:
   * read-subscripts.js (will be renamed read-dimensions.js or similar in a later branch)

---

Read variables phase:
* old:
   * variable-reader.js
* new:
   * read-variables.js

---

Read equations phase:
* old:
   * equation-reader.js
* new:
   * read-equations.js
   * read-equation-fn-*.js

---

Reduce equations phase:
* old:
   * expr-reader.js
* new:
   * reduce-variables.js (with support from `parse` package)

---

Generate code phase:
* old:
   * equation-gen.js
   * loop-index-vars.js
   * model-lhs-reader.js
   * var-name-reader.js
* new:
   * gen-equation.js (entrypoint that depends on the following implementation files)
      * direct-data-helpers.js
      * gen-const-list.js
      * gen-direct-const.js
      * gen-expr.js
      * gen-lookup-*.js

chrispcampbell:
Fixes #412 

@ToddFincannonEI: This PR adds the new reader and code gen implementations that I described earlier.  See issue for more info.  Let me know if you'd like more details on anything.

A couple notes about compatibility:

* Both old and new code are passing 100% of the unit tests in the `compile` package (the `.spec.ts` files I added recently).

* I added a new "compare" mode to the existing `modeltests` script.  So now if you do `pnpm test:c-int compare [model]` it will run through all models (or the specified one), and for each one it builds the model twice, once with the old and once with the new, and then compares the generated code (using `git diff -w` to ignore minor whitespace differences).  The new code passes the comparison test for all models except for 3 that have minor differences in code gen due to how parentheses are handled or how the `IF THEN ELSE` are reduced, but otherwise they produce the correct model outputs.

* I tested En-ROADS and the new code produces exactly the same model outputs.  There are only some slight differences in the generated code due to how parens are reduced and a few cases where an `IF THEN ELSE` reduces to a constant (in the old code it would be treated as an aux, in the new code it gets treated as a const, but otherwise they behave the same).

* I also tested C-ROADS ~and there's one small issue with the new code, which I'll either fix on this branch or save for later~.  (I forgot to finish implementing support for `specialSeparationDims` before I filed the PR.  I have since pushed a fix to this branch along with a test since we didn't already have coverage for it.)

* If you want to test EPS with the new code, use `SDE_NONPUBLIC_USE_NEW_PARSE=1 sde ...`.

* If you want a better apples-to-apples comparison of the generated code between old and new, I'd recommend setting `SDE_CODE_GEN_CHUNK_SIZE=0` which will disable the function chunking in code gen that makes diffs harder to interpret.

* If it's helpful, you can also disable the `IF THEN ELSE` reduction optimization for both old and new implementations using `SDE_NONPUBLIC_REDUCE_VARIABLES=0`.


chrispcampbell:
For the record, Todd tested the latest version of the EPS model against this branch and found that the new parsing + code gen paths produce results that are compatible with the old, so he is comfortable with merging this PR.  I will merge it to main shortly.

PR #423, opened 2024-01-11
fix: add new expandVarNames function that works with both old and new parsers

### LINKED ISSUE #422: Make `expandedVarNames` work with new parser
The `expandedVarNames` function used in the `code-gen` module currently relies on the `ModelLHSReader` class from the old parsing subsystem (which has a direct dependency on `antlr4-vensim`).  We need to add a new code path so that when the new parser is active, we use that instead of `ModelLHSReader`.

chrispcampbell:
Fixes #422 

@ToddFincannonEI: This adds a new function (in a separate module, so that it's easier to test) that is used by `expandedVarNames` in code gen, so that it works when either "old" or "new" reader code is active.

This is the last place (that I know of) where "old" reader code was being used when "new" reader is active.

Note the two TODO comments in the test file.  Since this function is only used in the case where a user is not supplying a `spec.json` file (with explicit lists of inputs and outputs), and since I expect most users will supply a `spec.json` file, I'm not too worried about making this function perfect right now.  The main thing is that the new implementation works mostly the same as the old one.


ToddFincannonEI (commented):
I didn't run the tests, since I know you will, but I read the code. I am looking forward to using the new parser the next time I update SDEverywhere for EPS. I can see it will be much more straightforward than the multitude of reader instances I was using.

chrispcampbell:
> I didn't run the tests, since I know you will, but I read the code.

Yep, the new tests are run automatically and are all passing.  Thanks for reviewing, will merge shortly.

PR #425, opened 2024-01-12
fix: allow `--genc` and `--list` arguments to be used in the same `sde generate` command

### LINKED ISSUE #424: Allow `--genc` and `--list` arguments to be used in the same `sde generate` command
Currently the `sde generate` command only allows for one "operation" to be performed at a time.  The build package therefore has to run `sde generate` twice (once with `--genc` then again with `--list`) which leads to redundant work.  It would be better if we could make the internal `parseAndGenerate` function take an array of operations instead of a single operation, which would allow for multiple operations to be performed in the same invocation.

chrispcampbell:
Fixes #424

See issue for more details.  This helps improve build performance by quite a bit when using the `sde bundle` or `sde dev` commands.

### DIFF
--- a/packages/build/src/build/impl/gen-model.ts
+++ b/packages/build/src/build/impl/gen-model.ts
@@ -163,19 +163,15 @@ async function flattenMdls(
 async function generateC(context: BuildContext, sdeDir: string, sdeCmdPath: string, prepDir: string): Promise<void> {
   log('verbose', '  Generating C code')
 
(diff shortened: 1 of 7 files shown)


ToddFincannonEI (approved):
This makes a lot of sense for the En-ROADS use case. We don't usually generate a listing for EPS, so I like the way that --genc and --list can be given separately or together instead of simply adding the listing to --genc.

PR #432, opened 2024-02-14
fix: require Node.js v18 as minimum supported version

### LINKED ISSUE #431: Update minimum supported Node.js version to v18
The README says that minimum Node.js version is v14, and our GitHub Actions workflows have been running on v16 for a long time, but both have reached end of life.

We should update the docs to suggest using either v18 or v20 as the minimum.  There's nothing in SDE yet that requires v20, but there was at least one place I saw this week where using v18 would be beneficial (to replace our use of the outdated byline package).

Since there might be some users who have reason to use an older version, I'd suggest using v18 as a minimum (and use that when building on GitHub Actions), and the README could recommend using v20 (latest LTS version).

This would also be a good time to add [engines](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines) to the `package.json` for at least the cli package (but maybe someday we should add it for all packages that depend on Node):
```json
{
  "engines": {
    "node": ">=18"
  }
}
```

chrispcampbell:
Fixes #431 

See issue for more details.  This updates the docs and build workflows to use Node v18 as minimum supported version.  I also added `engines` to the `package.json` for the cli package.  (This is more advisory; it only produces a warning if the current node version is < 18, unless `engine-strict` config is enabled.)

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -55,7 +55,7 @@ jobs:
       - name: Install Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: 16
+          node-version: 18
 
       # The pnpm caching strategy in the following steps is based on:
       #   https://github.com/pnpm/action-setup#use-cache-to-reduce-installation-time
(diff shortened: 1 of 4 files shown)


ToddFincannonEI (commented):
I think you covered everything. Recommending but not requiring Node LTS sounds about right.

PR #434, opened 2024-02-20
fix: improve error handling/reporting and prevent premature exit in dev mode

### LINKED ISSUE #260: Improve error handling in watch mode when dat file can't be read (or emcc fails)
The `build` package has error handling in place already so that errors are caught and a message is displayed appropriately. However, there are still a couple cases where the build process exits while in watch mode when some error is encountered, for example if a dat file can't be read, or emcc/clang fails. We need to fix those cases so that the build process stays running and the error messages are reported correctly.

chrispcampbell:
Fixes #260 

@ToddFincannonEI: The gory details are in the issue, but TL;DR this improves error handling in a number of areas to:
1. add error handling in places where we didn't have it quite right (like in `readDat`) and
2. avoid `process.exit` and instead allow errors to bubble up and get reported in a standard way

### DIFF
--- a/packages/build/src/build/build.ts
+++ b/packages/build/src/build/build.ts
@@ -73,16 +73,14 @@ export async function build(mode: BuildMode, options: BuildOptions): Promise<Res
   const overlayEnabled = mode === 'development'
   setOverlayFile(messagesPath, overlayEnabled)
 
-  // Initialize plugins
-  const plugins = userConfig.plugins || []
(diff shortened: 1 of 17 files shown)


ToddFincannonEI (commented):
Improved error handling is long overdue. I like what you've done here. I reviewed the changes in the compile package, since I am not that familiar with the new packages. Letting exceptions bubble up instead of calling process.exit() seems obvious in retrospect!

PR #439, opened 2024-03-06
fix: prevent subscripted input variables from being pruned

### LINKED ISSUE #438: Incorrect C code generated when subscripted input variable is not referenced elsewhere
If you have a model that specifies an input variable (in the model spec `inputVarNames`), and that variable has subscripts, AND that variable is not referenced (i.e., it would otherwise be pruned due to it not being in the dependency chain of an output variable), then SDE will generate C code that references that variable, but there won't be a correct declaration for the variable (because it is getting pruned prematurely), and the build will fail when compiling the C code.

This is a bit of an edge case, because in this scenario such a slider would have no impact on the model behavior, but it would still be good to correct the issue because it can be confusing for a user of SDE to encounter a compilation error (and there is no other guidance to explain why it is failing).

The fix is to change the `removeUnusedVariables` code to ignore the subscript part when processing `inputVars` just like we do for `outputVars`.  I will add an automated test case to cover this scenario.

Example `inputs.csv`:
```csv
2,slider,view1,example slider[cn],example slider,,Assumptions,1,10,5,1,years,,,,,,,,,,,,,,,,Description goes here
```

Error message:
```
Input file inputs.csv has been changed
Generating files...
Done generating files (0.0s)
Generating model...
    Generating WebAssembly module
    build/processed.c:13500:6: error: use of undeclared identifier '_example_slider'
     13500 |
    &_example_slider[2],
                 |            ^
    build/processed.c:13518:3: error: use of undeclared identifier '_example_slider'
     13518 |     _example_slider[2] = i
    nputData[1];
                 |     ^
    2 errors generated.
    emcc: error: '/Users/campbell/Projects/CI/emsdk/upstream/bin/clang -target wasm32-unknown-emscripten -fignore-exceptions -fvisibility=default -mllvm -combiner-global-alias-analysis=false -mllvm -enable-emscripten-sjlj -mllvm -disable-lsr -Werror=implicit-function-declaration --sysroot=/Users/campbell/Projects/CI/emsdk/upstream/emscripten/cache/sysroot -Xclang -iwithsysroot/include/fakesdl -Xclang -iwithsysroot/include/compat -Ibuild -Wall -Os build/processed.c -c -o /var/folders/8r/0kjp9h6s649gf7k_p66rd7mr0000gn/T/emscripten_temp_43ba8k3p/processed_0.o' failed (returned 1)

ERROR: Failed to compile C model to WebAssembly: emcc command failed (code=1)
        at buildWasm (file:///Users/campbell/Projects/CI/example/node_modules/@sdeverywhere/plugin-wasm/dist/index.js:99:11)
        at async WasmPlugin.postGenerateC (file:///Users/campbell/Projects/CI/example/node_modules/@sdeverywhere/plugin-wasm/dist/index.js:24:5)
        at async generateModel (file:///Users/campbell/Projects/CI/example/node_modules/@sdeverywhere/build/dist/index.js:518:18) 
```

chrispcampbell:
Fixes #438 

@ToddFincannonEI: Here's a fix for an issue that a newcomer to SDE encountered today.  It's kind of an edge case that doesn't matter much once you get a project up and running, but can be a source of confusion when starting a new project.  See issue for more details.

Here's the diff with expanded scope to make it easier to see that the only change is to apply the same handling for input variables that we use for output variables (chop off the subscript part):

image

### DIFF
--- a/packages/compile/src/generate/gen-code-c.spec.ts
+++ b/packages/compile/src/generate/gen-code-c.spec.ts
@@ -63,7 +63,68 @@ describe('generateCode (Vensim -> C)', () => {
       y = :NOT: x ~~|
     `
     const code = readInlineModelAndGenerateC(mdl)
-    expect(code).toMatch('#include "sde.h"')
+    expect(code).toMatch(`\
+#include "sde.h"
+
+// Model variables
+double _x;
+double _y;`)
+  })
+
+  it('should work when valid input variable name without subscript is provided in spec file', () => {
+    const mdl = `
+      x = 10 ~~|
+      y = x + 1 ~~|
+    `
+    const code = readInlineModelAndGenerateC(mdl, {
+      inputVarNames: ['x'],
+      outputVarNames: ['y']
+    })
+    expect(code).toMatch(`\
+#include "sde.h"
+
+// Model variables
+double _x;
+double _y;`)
+  })
+
+  it('should work when valid input variable name with subscript (referenced by output variable) is provided in spec file', () => {
+    const mdl = `
+      DimA: A1, A2 ~~|
+      A[DimA] = 10, 20 ~~|
+      B[DimA] = A[DimA] + 1 ~~|
+    `
+    const code = readInlineModelAndGenerateC(mdl, {
+      inputVarNames: ['A[A1]'],
+      outputVarNames: ['B[A1]', 'B[A2]']
+    })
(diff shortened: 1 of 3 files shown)


ToddFincannonEI (commented):
Makes sense. Glad to see tests coming with the changes — something I never did.

PR #443, opened 2024-03-08
fix: rewrite recordRefsOfVariable to avoid blowing the stack

### LINKED ISSUE #442: Model reader running out of stack with a large model
The latest EPS model is much larger than previous versions. The SDEverywhere code gen command often crashes after running out of stack.

```
Generating the C model...
RangeError: Maximum call stack size exceeded
    at node_modules/.pnpm/ramda@0.27.2/node_modules/ramda/src/internal/_curry3.js:38:31
    at f1 (node_modules/.pnpm/ramda@0.27.2/node_modules/ramda/src/internal/_curry1.js:19:17)
    at find (node_modules/.pnpm/ramda@0.27.2/node_modules/ramda/src/find.js:47:9)
    at node_modules/.pnpm/ramda@0.27.2/node_modules/ramda/src/internal/_dispatchable.js:50:15
    at Module.f2 (node_modules/.pnpm/ramda@0.27.2/node_modules/ramda/src/internal/_curry2.js:34:14)
    at findVarWithRefId (packages/compile/src/model/model.js:652:19)
    at varWithRefId (packages/compile/src/model/model.js:664:16)
    at recordRefsOfVariable (packages/compile/src/model/model.js:452:24)
    at recordRefsOfVariable (packages/compile/src/model/model.js:455:11)
    at recordRefsOfVariable (packages/compile/src/model/model.js:455:11)
```

ToddFincannonEI:
Fixes #442

I refactored the tail-recursive recordRefsOfVariable function as an iterative function with an auxiliary stack. Details in #442.

### DIFF
--- a/packages/compile/src/model/model.js
+++ b/packages/compile/src/model/model.js
@@ -395,14 +395,12 @@ function removeUnusedVariables(spec) {
   // ensures that we include all subscripts for a variable, which might mean we
   // include some subscripts that aren't needed, but it is safer than trying to
   // eliminate those and possibly omit something that is needed.
-  const referencedVarNames = []
+  const referencedVarNames = new Set()
 
   // Add the given variable name to the list of referenced variables, if it's not
   // already there.
   const recordUsedVarName = varName => {
-    if (!referencedVarNames.includes(varName)) {
-      referencedVarNames.push(varName)
-    }
+    referencedVarNames.add(varName)
   }
 
   // Add the given variable to the list of referenced variables, and do the same for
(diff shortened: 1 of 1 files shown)


chrispcampbell (approved):
Looks good, thanks for the fixes and performance improvements.

I verified that the code produces the same results as before for En-ROADS.

FWIW I quickly made an automated test for this.  With a deep dependency tree (~50000 equations deep), the test fails as expected before your fix, and gets farther with your fix applied.  But the test then fails due to blowing the stack in toposort, but that's a separate issue and apparently not affecting EPS yet so I won't worry about it for now.  I'll add the test to the repo after this PR is merged (but skipped by default since it's slow to run).

```ts
  // Note that this test takes a while (> 8 seconds) to run, so is skipped by default
  it.skip('should work without exceeding stack limits when model has deep dependency tree', () => {
    const n = 50000
    let mdl = 'x0 = 1 ~~|\n'
    for (let i = 1; i <= n; i++) {
      mdl += `x${i} = x${i - 1} + 1 ~~|`
    }

    const code = readInlineModelAndGenerateC(mdl, {
      inputVarNames: ['x0'],
      outputVarNames: [`x${n}`]
    })

    for (let i = 0; i <= n; i++) {
      expect(code).toContain(`double _x${i};`)
    }
  })
```

PR #446, opened 2024-03-11
fix: make new implementation of read and code gen phases the default

### LINKED ISSUE #445: Make new implementation of read and code gen phases the default
A new implementation of the parse/read/gen phases was added in #412, but was disabled by default and could only be enabled with a temporary `SDE_NONPUBLIC_USE_NEW_PARSE` environment variable.  We have been using this new code path for a while when building En-ROADS and C-ROADS, and Todd has verified that EPS builds/runs fine using the new code path, so I think we should enable the new implementation by default (make it opt-out instead of opt-in).

We can leave the old code in place for another month or so, in case any other projects run into an issue with the new code path.

chrispcampbell:
Fixes #445

@ToddFincannonEI: This flips the logic for the temporary `SDE_NONPUBLIC_USE_NEW_PARSE` flag so that it's opt-out instead of opt-in.  I kinda regret not adding a helper function instead of checking the env var values inline, but these conditionals will be removed soon so it's not worth it.  I've replaced all `=== '0'` with `!== '1'` and vice versa.


ToddFincannonEI (commented):
Pretty straightforward! Glad to see this move into production.

Issue #449, opened 2024-03-22
Error related to use of SMOOTH3I when compiling C model to WebAssembly


serman:
I am including all the variables that we need for our project in the outputs.csv file

But one of them gives us an error:
```
variable name
Average wellbeing index[af]
Average wellbeing index[cn]
Average wellbeing index[ec]
Average wellbeing index[eu]
Average wellbeing index[la]
Average wellbeing index[me]
Average wellbeing index[pa]
Average wellbeing index[sa]
Average wellbeing index[se]
Average wellbeing index[us]
```

The error I get is this one:

```
10:18:34 AM [vite] hmr update /src/overlay.ts?v=e77358aa
  build/processed.c:17037:47: error: invalid operands to binary expression ('double[10]' and 'double')
  __level123 = _INTEG(__level123, (__level122 - __level123) / (_time_to_smooth_sdg_scores_for_wellbeing / 3.0));
                                     ~~~~~~~~~~ ^ ~~~~~~~~~~
  build/vensim.h:27:41: note: expanded from macro '_INTEG'
  #define _INTEG(value, rate) ((value) + (rate) * _time_step)
                                          ^~~~
  1 error generated
10:18:35 AM [vite] hmr update /src/dev-overlay.js
  emcc: error: '/Users/sergiogalan/proyectos/earth4all/testm/emsdk/upstream/bin/clang -target wasm32-unknown-emscripten -fignore-exceptions -fno-inline-functions -fvisibility=default -mllvm -combiner-global-alias-analysis=false -mllvm -enable-emscripten-sjlj -mllvm -disable-lsr -D__EMSCRIPTEN_major__=2 -D__EMSCRIPTEN_minor__=0 -D__EMSCRIPTEN_tiny__=34 -D_LIBCPP_ABI_VERSION=2 -Werror=implicit-function-declaration -Xclang -iwithsysroot/include/SDL --sysroot=/Users/sergiogalan/proyectos/earth4all/testm/emsdk/upstream/emscripten/cache/sysroot -Xclang -iwithsysroot/include/compat -Ibuild -Wall -Os build/processed.c -c -o /var/folders/v7/6lvtsm551jgdfdwjm_j5m7mr0000gn/T/emscripten_temp_g_81h4jc/processed_0.o' failed (returned 1)

ERROR: Failed to compile C model to WebAssembly: emcc command failed (code=1)
[...]

PR #457, opened 2024-03-27
fix: correct generation of level variables when SMOOTH3[I] input param has subscripts but delay param does not

### LINKED ISSUE #313: Incorrect code generated when SMOOTH3I input param has subscripts but delay param does not
We ran into an SDE issue when using `SMOOTH3I` in the En-ROADS model.  The following changes can be applied to the `smooth` test to reproduce:

```diff
--- a/models/smooth/smooth.mdl
+++ b/models/smooth/smooth.mdl
@@ -13,6 +13,9 @@ delay = 2 ~~|
 delay 2[SubA] = 2 ~~|
 delay 3[DimA] = 2 ~~|
 
+initial s = 50 ~~|
+initial s with subscripts[DimA] = 50 ~~|
+
 s1 = SMOOTH(input, delay) ~~|
 s2[DimA] = SMOOTH(input, delay) ~~|
 s3[DimA] = SMOOTH(input 3[DimA], delay 3[DimA]) ~~|
@@ -23,6 +26,10 @@ s7[SubB] = SMOOTH(input 2[SubA], delay 2[SubA]) ~~|
 s8[DimA,DimB] = SMOOTH(input 3x3[DimA,DimB], delay) ~~|
 s9[SubA,DimB] = SMOOTH(input 2x3[SubA,DimB], delay) ~~|
 s10[SubA,B1] = SMOOTH(input 2[SubA], delay) ~~|
+s11[DimA] = SMOOTH3(input 3[DimA], delay) ~~|
+s12[DimA] = SMOOTH3I(input 3[DimA], delay 3[DimA], initial s) ~~|
+s13[DimA] = SMOOTH3I(input 3[DimA], delay, initial s) ~~|
+s14[DimA] = SMOOTH3I(input 3[DimA], delay, initial s with subscripts[DimA]) ~~|
 
 INITIAL TIME  = 0 ~~|
 FINAL TIME  = 40 ~~|
```

The above model works fine in Vensim, but in SDE, the `s13` equation causes incorrect code generation:

```
$ pnpm test:c-int smooth

> sdeverywhere-monorepo@ test:c-int /Users/campbell/Projects/CI/sdeverywhere
> ./tests/modeltests "smooth"

Testing the smooth model
smooth.c:443:41: error: invalid operands to binary expression ('double[3]' and 'double')
  __level9 = _INTEG(__level9, (__level8 - __level9) / (_delay / 3.0));
                               ~~~~~~~~ ^ ~~~~~~~~
./vensim.h:23:41: note: expanded from macro '_INTEG'
#define _INTEG(value, rate) ((value) + (rate) * _time_step)
                                        ^~~~
1 error generated.
make: *** [smooth.o] Error 1
```

The bug occurs when the first (input) parameter is subscripted, but the second (delay) parameter is not.  It looks like the `generateSmoothLevel` code in `equation-reader.js` might be making incorrect assumptions about the presence of subscripts in the different parameters.  I didn't dig in deeper to diagnose.

The workaround is to change the variable used for the second parameter to include a compatible dimension, even though it would normally not be necessary.

chrispcampbell:
Fixes #313 

This fixes an issue that we ran into with En-ROADS, but was also reported last week in #449 (will be closed as a duplicate).  See #313 for more details on the cause and the fix.

To help verify this fix, I've added new tests for the equation reading phase (when the level vars are synthesized for the implementation of `SMOOTH3[I]`) and for the code gen phase.

I also verified that this fixes the issue reported in the duplicate issue #449.

I decided to only fix this in the "new" `readEquations` code, since the old `EquationReader` is no longer enabled by default and will be removed in the near future.


ToddFincannonEI (commented):
This fix looks like a straightforward solution that should not cause any problems for EPS.

PR #462, opened 2024-04-18
fix: use original Vensim names in generated headers for spec output vars

### LINKED ISSUE #461: `sde compare` fails to detect output errors
The attached test model illustrates a case where the `sde compare` command fails to detect errors in C model output. It only occurs when a spec file gives output variables and subscripts are reordered.

1. Run `validation-subs.mdl` in Vensim and export to dat.
2. Run `sde generate --genc --spec spec.json validation-subs`
3. Run `sde compile validation-subs`
4. Run `sde exec validation-subs`
5. Run `sde log --dat output/validation-subs.txt`
6. Manually edit the `output/validation-subs.txt` file so that `a[B1,A1]` has a value of 0 at time 0.
7. Run `sde compare validation-subs.dat output/validation-subs.dat`. No errors are detected.

[validation-subs.zip](https://github.com/climateinteractive/SDEverywhere/files/15027390/validation-subs.zip)

ToddFincannonEI:
Fixes #461

See issue #461 for details.


packages/compile/src/generate/code-gen.js
```diff
@@ -157,15 +157,15 @@ ${chunkedFunctions('evalLevels', Model.levelVars(), '  // Evaluate levels.')}
...
+    let headerVars = outputAllVars ? expandedVarNames(true) : spec.outputVarNames
     let outputVars = outputAllVars ? expandedVarNames() : spec.outputVars
     mode = 'io'
     return `void setInputs(const char* inputData) {${inputsFromStringImpl()}}
 
 void setInputsFromBuffer(double* inputData) {${inputsFromBufferImpl()}}
 
 const char* getHeader() {
-  return "${R.map(varName => headerTitle(varName), headerVars).join('\\t')}";
+  return "${headerVars.join('\\t')}";
```
chrispcampbell, line 168: @ToddFincannonEI: It looks like the build is failing due to a test failure when running the "allocate" model, but probably others are affected as well:
https://github.com/climateinteractive/SDEverywhere/actions/runs/8741036880/job/23986139008

The error points back to this line, which will fail if `headerVars` is undefined.

The issue is that there are two ways to specify output variables in the `spec.json`, you can either use `outputVars` (which takes C names) or `outputVarNames` (which takes Vensim names).  In the case of `outputVarNames`, we convert to `outputVars` in this code in `Model.read` so that the rest of the code only needs to deal with `outputVars` (sort of a lowest common denominator):

```
  if (spec) {
    // If the spec file contains `input/outputVarNames` (with full Vensim variable names)
    // convert those to C names first.  Otherwise, use `input/outputNames` which are already
    // assumed to be valid C names.
    if (spec.inputVarNames) {
      spec.inputVars = R.map(cName, spec.inputVarNames)
    }
    if (spec.outputVarNames) {
      spec.outputVars = R.map(cName, spec.outputVarNames)
    }
    // Save the input vars locally so that they can be referenced by `isInputVar`.
    if (spec.inputVars) {
      inputVars = spec.inputVars
    }
  }
```

There are only 3 sample models that use the older `outputVars` form (allocate, arrays_cname, longeqns).

There are a few different ways we could solve this.  I don't think it's imperative that we retain support for `outputVars`.  For En-ROADS/C-ROADS we use the newer `outputVarNames` and all the new build stuff in SDE also uses that form, so I doubt anyone out in the wild is relying on `outputVars`.  I think if we changed those 3 tests to use `outputVarNames` and effectively drop support for `outputVars`, then your proposed fix will be nearly correct (though I'd need to look through the rest of the code to see if any other changes are needed).

Let me know which approach you'd like to take.

ToddFincannonEI: Sorry I forgot to run the tests before submitting. EPS uses outputVarNames too, so your proposed solution looks good to me. I will change the tests and resubmit.

ToddFincannonEI: It looks like the only purpose of the arrays_cname test model is to test the `outputVars` C name support in spec files. I propose deleting this model, since we are dropping support for `outputVars`.

chrispcampbell: @ToddFincannonEI: I agree with deleting arrays_cname, and also suggest renaming arrays_varname back to just arrays, but I'd be OK handling that myself in a separate PR after this one.

ToddFincannonEI:
@chrispcampbell I made the changes to the test models we discussed and found one other bug in the process. It's ready for you to review again.

packages/compile/src/generate/code-gen.js
```diff
@@ -17,14 +17,7 @@ let codeGenerator = (parsedModel, opts) => {
...
   // Set to true to output all variables when there is no model run spec.
-  let outputAllVars
-  if (spec.outputVars && spec.outputVars.length > 0) {
-    outputAllVars = false
-  } else if (spec.outputVarNames && spec.outputVarNames.length > 0) {
-    outputAllVars = false
-  } else {
-    outputAllVars = true
-  }
+  let outputAllVars = spec.outputVarNames === undefined || spec.outputVarNames.length === 0
```
chrispcampbell, line 20: I simplified this logic now that the user can only provide `outputVarNames` and `outputVars` is only used internally.

packages/compile/src/generate/code-gen.js
```diff
@@ -157,19 +150,19 @@ ${chunkedFunctions('evalLevels', Model.levelVars(), '  // Evaluate levels.')}
   // Input/output section
   //
   function emitIOCode() {
-    let headerVars = outputAllVars ? expandedVarNames(true) : spec.outputVars
-    let outputVars = outputAllVars ? expandedVarNames() : spec.outputVars
+    let headerVarNames = outputAllVars ? expandedVarNames(true) : spec.outputVarNames
+    let outputVarIds = outputAllVars ? expandedVarNames() : spec.outputVars
```
chrispcampbell, line 154: I renamed these two variables to emphasize the difference, but otherwise it's the same as you had it.

ToddFincannonEI (commented):
Thanks for finding all the other references to outputVars and clarifying the other references to outputVarNames. Much improved!

packages/compile/src/model/model.js
```diff
@@ -102,9 +102,8 @@ function read(parsedModel, spec, extData, directData, modelDirname, opts) {
   if (opts?.stopAfterReadVariables) return
 
   if (spec) {
-    // If the spec file contains `input/outputVarNames` (with full Vensim variable names)
-    // convert those to C names first.  Otherwise, use `input/outputNames` which are already
-    // assumed to be valid C names.
+    // If the spec file contains `input/outputVarNames`, convert the full Vensim variable
+    // names to C names first so that later phases only need to work with canonical names
```
chrispcampbell, line 106: I updated this comment to make it match the new reality better (also `input/outputNames` had the wrong property names anyway).

packages/cli/README.md
```diff
@@ -188,8 +188,8 @@ Be sure to include `Time` first among the output variables.
 
 ```json
 {
-  "inputVars": ["Reference predators", "Reference prey"],
-  "outputVars": ["Time", "Predators Y", "Prey X"]
+  "inputVarNames": ["Reference predators", "Reference prey"],
+  "outputVarNames": ["Time", "Predators Y", "Prey X"]
```
chrispcampbell, line 192: I updated the README (which I guess had been incorrect for a while) so that it only shows using `input/outputVarNames`.

chrispcampbell (approved):
@ToddFincannonEI: Looks good, thanks, Todd.  I made a couple small tweaks to comments/docs which I noted in inline comments, the rest looked good so I will merge now.

PR #465, opened 2024-04-19
fix: upgrade to more recent version of xlsx package

### LINKED ISSUE #463: Upgrade to more recent version of xlsx package
We are currently using v0.17.5 of the xlsx (aka SheetJS) package.  That package has had a few security alerts in the past couple years, the latest one says to upgrade to v0.20.2.  One hitch is that the SheetJS maintainers have been in a battle with npmjs for a while and no longer publish their packages to the npm registry, so their instructions now say to install from their own CDN:

```
    "xlsx": "https://cdn.sheetjs.com/xlsx-0.20.2/xlsx-0.20.2.tgz"
```

I'm not a fan of relying on a self-hosted binary (which they could take down at any time), and I'm not a fan of their development approach, but unfortunately there are no easy alternatives at this time.  The other ones I looked into, exceljs and read-excel-file, only support async reading of files, so even if I wanted to try them, I'd have to change a lot of compiler functions (on the code gen side) to be async ready, which is not a quick task.  So for now, I'm afraid the best solution is to point to the most recent version of their self-hosted binary.

(I would still like to replace this dependency someday, maybe as part of changing the compiler to read/cache the xlsx/csv data in an earlier step rather than waiting till the code gen phase.)

chrispcampbell:
Fixes #463 

@ToddFincannonEI: This upgrades the xlsx package to a more recent version, which addresses some security vulnerabilities that have been reported.  (Those vulnerabilities cause scary warnings when a user does an npm install of `@sdeverywhere/cli`, so it would be good to avoid those warnings, which can be a turn-off for new users.)  I have misgivings about this package and using their CDN (see issue comments), but think that this should be a relatively simple upgrade as compared to replacing the package outright.

### DIFF
--- a/packages/compile/package.json
+++ b/packages/compile/package.json
@@ -25,7 +25,7 @@
     "js-yaml": "^3.13.1",
     "ramda": "^0.27.0",
     "split-string": "^6.0.0",
-    "xlsx": "^0.17.0"
+    "xlsx": "https://cdn.sheetjs.com/xlsx-0.20.2/xlsx-0.20.2.tgz"
   },
   "author": "Climate Interactive",
   "license": "MIT",
--- a/packages/compile/src/_shared/helpers.js
+++ b/packages/compile/src/_shared/helpers.js
@@ -1,3 +1,4 @@
+import * as fs from 'node:fs'
 import util from 'util'
 import B from 'bufx'
 import { parse as parseCsv } from 'csv-parse/sync'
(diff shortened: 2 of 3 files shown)


ToddFincannonEI (approved):
I can't add to the tests. EPS does not read any XLSX files, only CSV. Looks good to me based on passing our other tests. I am fine with using a non-npm package source when there is no other choice. This package has worked well for me for other utilities, so I don't see a good reason to switch. Open to alternatives though.

PR #490, opened 2024-05-31
feat: allow for overriding data variables and lookups at runtime

### LINKED ISSUE #472: Allow for overriding data variables and lookups at runtime
This feature was originally filed together with adding support for overriding constant variables (see #470) but I now think it would be easier to manage as a separate issue, so I'm filing this as a standalone feature.

To enable rapid prototyping and calibration of our models, we would like to have a way to override data or lookup variables at runtime.

I prototyped this in my work-in-progress "generate JS" branch because I needed it for a small model we were working on. I believe that @ToddFincannonEI already has a related branch, so we should put our heads together on this.

chrispcampbell:
Fixes #472

This updates the compiler so that it includes a `setLookup` function in generated C and JS models, and then updates the runtime packages to allow for supplying lookup overrides via a new `options` parameter for `ModelRunner.runModel`.

See this [issue comment](https://github.com/climateinteractive/SDEverywhere/issues/472#issuecomment-2141173382) for more details on the changes included here.  TL;DR this mostly adds new functionality, and there should be no impact on existing projects.

/cc @ToddFincannonEI


chrispcampbell:
I resolved one last minute issue with the runtime-async tests, otherwise I think this is ready to be merged.  As with the other recent branches and PRs, Todd won't have time to review this before it gets merged, so we will merge but then keep the changes queued up on main before we publish official packages.

PR #596, opened 2025-01-09
fix: increase C input buffer size + change default optimization flag for C compilation to `-O2`

### LINKED ISSUE #595: Update the C executable to support EPS 4
The latest EPS 4 model from Energy Innovation is much larger than previous versions. We generate a C executable as part of the validation process that compares SDEverywhere model output with Vensim output.

Validation of reference scenarios requires an input file for the C executable containing constant and lookup overrides. This goes into a static buffer. We need to increase the size of this buffer to accommodate our latest reference scenarios. A more flexible approach would be dynamically allocate the required memory, but that seems like overkill for this infrequently-used feature.

The EPS web app uses the -O2 optimization level in Emscripten, based on tests that balance performance with wasm file size. The C makefile needs to change to use -O2 to ensure matching results. We have run into cases where different optimization settings can results in output differences. This should not happen, but in lieu of investigating the problem, I have opted to simply match the optimization level for the C executable.

ToddFincannonEI:
Fixes #595

See issue #595 for details.

### DIFF
--- a/packages/cli/src/c/main.c
+++ b/packages/cli/src/c/main.c
@@ -2,7 +2,7 @@
 
 int main(int argc, char** argv) {
   // TODO make the input buffer size dynamic
-  char inputs[1500];
+  char inputs[500000];
   // When true, output data without newlines or a header, suitable for embedding reference data.
   bool raw_output = false;
   // When true, suppress data output when using PR* macros.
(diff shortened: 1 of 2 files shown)


chrispcampbell (approved):
This looks fine to me.  At CI we don't generally rely on the compiled native executable for much, and this change is not likely to affect many users in any case, so seems OK.

PR #602, opened 2025-02-05
feat: add `separateAllVarsWithDims` option to spec file to separate all vars with a dimension on the LHS

### LINKED ISSUE #601: The `specialSeparationDims` list is hard to maintain
EPS has an ongoing need to use the `specialSeparationDims` section in the spec file. The current format of this section requires listing every variable that needs to be separated to avoid a cyclic dependency condition. Maintaining that list is an error-prone process, because affected variables come and go as the model is developed.

ToddFincannonEI:
Fixes #601 

This PR proposes an alternative to the `specialSeparationDims` syntax to avoid having to maintain the list in cases where all variables containing a given subscript should be separated.

The spec file can now contain a property named `separateAllVarsWithDims` with an array value. Each entry is either a dimension name string in canonical format, or an array of dimension names.

If a variable LHS matches all dimension names in one of the `separateAllVarsWithDims` entries, then it is separated on those dimensions. If a variable is specifically called out in   `specialSeparationDims`, then that takes precedence, and `separateAllVarsWithDims` is not checked. It does not make sense for a variable separation to be listed in both formats.

I added a test model called `separateall.mdl` that has two apply-to-all arrays named `a` and `b`.

```
DimA: A1, A2 ~~|
DimB: B1, B2 ~~|
a[DimA] = c ~~|
b[DimB] = c ~~|
c = 42 ~~|
```

Without a spec file, the following code and variable listing is generated.

```
// a[DimA] = c
for (size_t i = 0; i < 2; i++) {
_a[i] = _c;
}
// b[DimB] = c
for (size_t i = 0; i < 2; i++) {
_b[i] = _c;

a[DimA]: aux
= c
refId(_a)
families(_dima)
subscripts(_dima)
hasInitValue(false)
refs(_c)

b[DimB]: aux
= c
refId(_b)
families(_dimb)
subscripts(_dimb)
hasInitValue(false)
refs(_c)
```

I made a spec file that tests both separation methods:
```
"specialSeparationDims": { "_a": "_dima" },
"separateAllVarsWithDims": ["_dimb"]
```

The variables are separated in both cases:
```
// a[DimA] = c
_a[0] = _c;
// a[DimA] = c
_a[1] = _c;
// b[DimB] = c
_b[0] = _c;
// b[DimB] = c
_b[1] = _c;

a[DimA]: aux (non-apply-to-all)
= c
refId(_a[_a1])
families(_dima)
subscripts(_a1)
separationDims(_dima)
hasInitValue(false)
refs(_c)

a[DimA]: aux (non-apply-to-all)
= c
refId(_a[_a2])
families(_dima)
subscripts(_a2)
separationDims(_dima)
hasInitValue(false)
refs(_c)

b[DimB]: aux (non-apply-to-all)
= c
refId(_b[_b1])
families(_dimb)
subscripts(_b1)
separationDims(_dimb)
hasInitValue(false)
refs(_c)

b[DimB]: aux (non-apply-to-all)
= c
refId(_b[_b2])
families(_dimb)
subscripts(_b2)
separationDims(_dimb)
hasInitValue(false)
refs(_c)
```

Because generation and execution do not fail in this test model, I added a check script that generates the model listing and confirms that both variables are separated. This is the only way to test the PR, since we don't have a failing test otherwise.

I tested this change on an EPS 4 model and confirmed that the new spec file format results in the same C code generation as the previous spec file that listed each variable separately.

### DIFF
--- a/packages/compile/src/model/read-equations.js
+++ b/packages/compile/src/model/read-equations.js
@@ -114,6 +114,9 @@ class Context {
     const parsedModel = { kind: 'vensim', root: parseVensimModel(eqnText) }
 
     // Create one or more `Variable` instances from the equations
+    // TODO: Technically we should pass `spec.separateAllVarsWithDims` here so that any
+    // generated apply-to-all variables will be separated according to the spec like other
+    // normal variables
     const vars = readVariables(parsedModel)
 
     // Add the variables to the `Model`
--- a/packages/compile/src/model/read-equations.spec.ts
+++ b/packages/compile/src/model/read-equations.spec.ts
@@ -25,6 +25,7 @@ function readSubscriptsAndEquationsFromSource(
   },
   opts?: {
     specialSeparationDims?: { [key: string]: string }
+    separateAllVarsWithDims?: string[][]
   }
 ): Variable[] {
   // XXX: These steps are needed due to subs/dims and variables being in module-level storage
@@ -40,7 +41,8 @@ function readSubscriptsAndEquationsFromSource(
   }
 
   const spec = {
-    specialSeparationDims: opts?.specialSeparationDims
+    specialSeparationDims: opts?.specialSeparationDims,
+    separateAllVarsWithDims: opts?.separateAllVarsWithDims
   }
 
   let modelDir = source.modelDir
@@ -67,6 +69,7 @@ function readInlineModel(
   modelDir?: string,
   opts?: {
     specialSeparationDims?: { [key: string]: string }
+    separateAllVarsWithDims?: string[][]
   }
 ): Variable[] {
   const vars = readSubscriptsAndEquationsFromSource({ modelText, modelDir }, opts)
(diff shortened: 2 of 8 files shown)


packages/compile/src/model/read-equations.spec.ts
```diff
@@ -192,6 +195,120 @@ describe('readEquations', () => {
     ])
   })
 
+  it('should work for equations that are affected by separateAllVarsWithDims', () => {
```
chrispcampbell, line 198: We had an existing unit test (or something closer to a unit test) for `specialSeparationDims` and I think we should do the same for this new `separateAllVarsWithDims` so I added them.  I always prefer to have vitest tests defined for things as the iteration cycle is much tighter than having to rely on sample model tests.

packages/compile/src/model/read-equations.spec.ts
```diff
@@ -192,6 +195,120 @@ describe('readEquations', () => {
...
+      DimA: A1, A2 ~~|
+      DimB: B1, B2 ~~|
+      DimC: C1, C2 ~~|
+      x[DimA] = 0 ~~|
+      y[DimB] = 0 ~~|
+      z[DimA, DimB, DimC] = 0 ~~|
+    `,
+      undefined,
+      {
+        separateAllVarsWithDims: [['_dima', '_dimc'], ['_dimb']]
```
chrispcampbell, line 210: In adding this test I noticed that the current implementation has a potential issue.  If you were to put the `['_dimb']` first here, then the implementation would hit that case first and would end up separating `z` over `DimB`.

I suppose that this case may not come up often in practice (and probably EPS will be the only user of this option for some time), but I wanted to flag it.  I guess that you can work around it by changing the order of the list like I did here, or falling back to using `specialSeparationDims` for more precise control.

packages/compile/src/model/read-variables.js
```diff
@@ -12,17 +12,21 @@ import Variable from './variable.js'
  *
  * @param {*} parsedModel TODO: Use ParsedVensimModel type here
  * @param {Object.<string, string>} [specialSeparationDims] The variable names that need to be
- * separated because of circular references.  A mapping from "C" variable name to "C" dimension
- * name to separate on.
+ * separated because of circular references.  This is an optional mapping from "C" variable name
+ * to "C" dimension name to separate on.
+ * @param {string[][]} [separateAllVarsWithDims] An optional array containing arrays of dimension
```
chrispcampbell, line 17: I changed the type to `string[][]` to make it slightly more obvious that how this works in the case of providing more than one dimension.

packages/compile/src/model/read-variables.js
```diff
@@ -115,6 +121,23 @@ function variablesForEquation(eqn, specialSeparationDims) {
...
+      // Alternatively, if the variable was not in `specialSeparationDims`, separate
+      // on dims from `separateAllVarsWithDims` if the var matches one of the dim lists.
+      if (separationDims.length === 0) {
+        for (let dimList of separateAllVarsWithDims) {
+          // The list entry from the spec is only considered a match if every dimension
+          // in the spec list appears on the LHS
+          if (dimList.every(dim => subIds.includes(dim))) {
+            // Technically we allow each entry in the spec array to be either a single
+            // dim ID string or an array of dim IDs
+            if (!Array.isArray(dimList)) {
```
chrispcampbell, line 133: I think it would be better to avoid having the implicit "up convert string to string[]" code so that you always have to provide a `string[][]` in the spec file, but I didn't make any changes to the code.  I guess I'm fine leaving it as is given that this is basically a hidden feature.

packages/compile/src/model/read-equations.js
```diff
@@ -114,6 +114,9 @@ class Context {
     const parsedModel = { kind: 'vensim', root: parseVensimModel(eqnText) }
 
     // Create one or more `Variable` instances from the equations
+    // TODO: Technically we should pass `spec.separateAllVarsWithDims` here so that any
```
chrispcampbell, line 117: I added a TODO here because it's possible that you could end up with generated vars that are not separated as expected.  In practice I don't think this is likely to happen given that `defineVariables` is typically used in cases where the variables are already non-apply-to-all, but I could be wrong about that.  Anyway, I think it's fine to not act on this now but wanted to flag it in case you think we should fix it now (it wouldn't be too hard to pass the `spec` to this function).

chrispcampbell (commented):
@ToddFincannonEI: I pushed some additional tests and comment changes, and added some inline comments here to call some things to your attention.  Given that this is kind of a hidden feature for now and probably only used by EPS, it's probably good enough but will let you decide.

The main thing I want to hear back from you is about the case of variables with more than one dimension; the code seems to be set up to accept that case but there wasn't a test for this other than the small one I added so I want to make sure the way I tested it is how you expected it to work, or if there are other cases we should add.  I'll wait for your feedback before I mark it approved.

packages/compile/src/model/read-variables.js
ToddFincannonEI: I was following the implementation we had for `specialSeparationDims`. I agree it would be safer to force the type, but in practice, it's nice to have the single string option.

packages/compile/src/model/read-equations.spec.ts
ToddFincannonEI: These are better tests than the model test I added. I should start doing it that way. Shall we remove the model test, since the unit test covers the same thing?

The tests work the way I expected. I am thinking of `separateAllVarsWithDims` as a kind of shorthand for `specialSeparationDims`. They both have the effect of setting `separationDims` for the `subscriptPositionsToExpand` call, so the dim list has the same semantics.

chrispcampbell: > Shall we remove the model test, since the unit test covers the same thing?

No, I think there's value in having both kinds of tests.  The model test helps verify things end-to-end (like, that the spec file is passed through the chain as expected).

chrispcampbell (approved):
Based on the discussion in the comments, I think this is fine to merge in its current state.  As noted, I think there's value in having both kinds of tests, and no urgent changes needed in the code, so I will merge this now and publish a release.

PR #645, opened 2025-08-27
build: upgrade to pnpm 10

### LINKED ISSUE #644: Upgrade to pnpm 10
pnpm 10 has been out for about 8 months so it would be a good time to upgrade.  There was a new lock file format in v9.  It looks like I will need to upgrade to v9 first so that the lock file is migrated, and then upgrade to v10.  (I first tried to upgrade straight to v10 but the dependencies in the lock file were upgraded, which is something we want to avoid.)

We last upgraded pnpm in 2023 (to v8) in #339.

chrispcampbell:
Fixes #644 

@ToddFincannonEI: I finished upgrading all of our internal repos to pnpm 10, and I've upgraded the SDEverywhere repo here as well.

The changes are pretty straightforward.  I upgrade to v9 first to ensure that the lock file was migrated to the v9 format without affecting any of the pinned dependency versions.  Once that was done, I upgraded to v10 and updated the settings noted below.

The v10 specific changes were:
* add "onlyBuiltDependencies" and "ignoredBuiltDependencies" to include/exclude the dependencies that should be built on install (new feature/requirement in v10)
* move the `linkWorkspacePackages` setting from `.npmrc` to `.pnpm-workspace.yaml` (the old location would still work, but I think the pnpm-specific config file is a better location and seems to be the preferred file as of v10)

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,8 +1,0 @@
-# Force pnpm to link local packages:
-#   https://pnpm.io/npmrc#link-workspace-packages
-# The default value of this option was true in pnpm <= 8, but was changed
-# to false in pnpm 9; setting it to true here should work for both versions.
-# Note that we don't use `workspace:` protocol for intra-repo dependencies
-# because those don't always play well with release-please, so instead we
-# use normal semver ranges (e.g., "^1.2.3") and let pnpm handle resolution.
-link-workspace-packages=true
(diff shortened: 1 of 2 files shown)


ToddFincannonEI (approved):
This all looks good on inspection. I updated my local SDEverywhere and confirmed the build still works.

PR #681, opened 2025-10-10
feat: include grouped and sorted arrays of variable instances in JSON listing

### LINKED ISSUE #534: Include grouped and sorted arrays of variable instances in JSON listing
As mentioned in #307, when diagnosing differences between in output between SDEverywhere and Vensim, it's helpful to be able to compare variables in the order that they are evaluated in the code generated by the SDEverywhere compiler.

Currently there are two limitations that make it difficult to do this sort of comparison:
* The JSON listing contains variables in roughly the order of evaluation, but they are not exactly grouped according to the exact order of evaluation.  Ideally we would have the variable instances grouped by type.
* It is necessary to get the "subscript indices" needed to access these internal variables at runtime, but currently the only way to get these is to do some complex expansion of the `dimensions` and `variables` objects to derive the correct subscript indices.  Ideally we would expose the specific details for each variable instance so that no post-processing is needed to access the information.

To address these limitations, I'm going to add a new `varInstances` field to the JSON listing that includes grouped arrays that are in evaluation order, one object per variable "instance", to make it relatively easy to request the data at runtime for a specific subscripted instance of a variable:

```json
  "varInstances": {
    "constants": [],
    "lookupVars": [],
    "dataVars": [],
    "initVars": [],
    "levelVars": [],
    "auxVars": []
  }
```

Here's an example item:

```json
      {
        "varId": "_g[_a1,_b2]",
        "varName": "g[A1,B2]",
        "varType": "const",
        "varIndex": 7,
        "subIndices": [
          0,
          1
        ]
      }
```

These groupings match the order of execution/evaluation in the generated model:
* initConstants (vars of type `const` only, called for t=0 only)
* initLookups (vars of type `lookup` only, called for t=0 only)
* initData (vars of type `data` only, called for t=0 only)
* initLevels (vars returned by `initVars`, a mix of initial, aux, and level vars, called for t=0 only)
* evalLevels (vars of type `level` only; called before `evalAux` for t>0)
* evalAux (vars of type `aux` only)

chrispcampbell:
Fixes #534

@ToddFincannonEI: I'm requesting your review on this one since it's in the core of the compiler and I wanted to bring it to your attention.  It doesn't have any impact on the behavior of the compiler and doesn't affect the existing JSON listing format (it only adds a new object to that listing file).

For reference, I already merged a couple small refactoring PRs that made these changes easier, see #678 and #680.  They were low-risk and well covered by tests so I figured I'd merge those first and send this one for your review to make it easier to see the distinct sets of changes.

### DIFF
--- a/packages/compile/src/model/model.js
+++ b/packages/compile/src/model/model.js
@@ -2,7 +2,7 @@ import B from 'bufx'
 import yaml from 'js-yaml'
 import * as R from 'ramda'
 
-import { decanonicalize, isIterable, strlist, vlog, vsort } from '../_shared/helpers.js'
+import { canonicalVensimName, decanonicalize, isIterable, strlist, vlog, vsort } from '../_shared/helpers.js'
 import {
   addIndex,
   allAliases,
@@ -15,6 +15,7 @@ import {
 } from '../_shared/subscript.js'
 import { cName } from '../_shared/var-names.js'
 
+import { expandVar } from './expand-var-instances.js'
 import { readEquation } from './read-equations.js'
 import { readDimensionDefs } from './read-subscripts.js'
 import { readVariables } from './read-variables.js'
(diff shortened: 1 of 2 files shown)


ToddFincannonEI (approved):
The code looks fine to me. I can't run it right now, because I'm in the middle of debugging a cyclic dependency in the latest EPS model, and I don't want to disturb my local working copy. This looks like it will be very helpful for debugging once I can merge it.

PR #772, opened 2026-02-10
test: update `delay` test models to use TIME STEP of 0.25 instead of 1

### LINKED ISSUE #771: Use smaller TIME STEP in `delay` test model
I've been trying to wrap up initial support for XMILE/Stella models in #468.  Most of the example models that were converted from Vensim mdl to Stella stmx in #517 are now passing on my `chris/468-xmile` branch.

However, I found that SDE was producing different results for `delay.stmx` than the ones produced by Stella (saved in `delay.csv`) for two variables:

```xml
			<aux name="d8">
				<dimensions>
					<dim name="DimA"/>
				</dimensions>
				<eqn>DELAY3(input, delay_a[DimA])</eqn>
				<isee:delay_aux/>
			</aux>

			<aux name="d11">
				<dimensions>
					<dim name="DimA"/>
				</dimensions>
				<eqn>k*DELAY3(input, delay_a[DimA])</eqn>
			</aux>
```

What's notable about these two is that they're testing `DELAY3`.  In the Vensim results (in `delay.dat`), the values for, say, `d8[A1]` would oscillate wildly around zero at t>7:

```
5	10
6	10
7	-260
8	1360
9	-5120
10	16480
```

SDE produces results that are compatible with Vensim.  But in the Stella results, these values would be zero, which caused the SDE model test to fail.

I eventually realized that the "delay" parameter values for the `DELAY3` function must be chosen carefully relative to the `TIME STEP` (or `dt` in XMILE notation), in particular the delay time should not be less than `TIME STEP`.  (The Vensim docs for [DELAY FIXED](https://www.vensim.com/documentation/fn_delay_fixed.html) and [DELAY N](https://www.vensim.com/documentation/fn_delay_n.html) allude to this.)

Since the delay value is effectively divided by 3 in the case of `DELAY3`, we need to make sure that the `TIME STEP` value is smaller than the delay time values.  In the case of these test models, the delay times vary from 1 to 3, and the `TIME STEP` was set to 1.

I propose that we change the `TIME STEP` in the mdl (and `dt` in the stmx) to 0.25 (which is a more realistic value anyway) but otherwise keeping the model equations the same, then generate new dat/csv files.

Once I made this change, Stella now produces nearly identical results as Vensim, and SDE also matches, so the `delay` test model passes for both Vensim and XMILE backends.

chrispcampbell:
Fixes #771 

@ToddFincannonEI: See issue for details.  This makes the `delay` models use more reasonable values for the `TIME STEP` and delay time parameters so that we can properly test the new XMILE support in SDE.

### DIFF
--- a/packages/compile/src/model/read-equations.spec.ts
+++ b/packages/compile/src/model/read-equations.spec.ts
@@ -5619,13 +5619,13 @@ describe('readEquations', () => {
         refId: '_final_time',
         varType: 'const'
       }),
-      v('TIME STEP', '1', {
+      v('TIME STEP', '0.25', {
         refId: '_time_step',
         varType: 'const'
       }),
-      v('SAVEPER', 'TIME STEP', {
+      v('SAVEPER', '1', {
         refId: '_saveper',
-        references: ['_time_step']
+        varType: 'const'
       }),
       v('Time', '', {
         refId: '_time',
--- a/packages/compile/src/model/read-variables.spec.ts
+++ b/packages/compile/src/model/read-variables.spec.ts
@@ -502,8 +502,8 @@ describe('readVariables', () => {
       }),
       v('INITIAL TIME', '0'),
       v('FINAL TIME', '10'),
-      v('TIME STEP', '1'),
-      v('SAVEPER', 'TIME STEP'),
+      v('TIME STEP', '0.25'),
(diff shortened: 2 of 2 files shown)


ToddFincannonEI (approved):
Looks good to me.

PR #798, opened 2026-04-08
fix: handle more special characters when converting variable names to canonical IDs

### LINKED ISSUE #794: Some special characters (math operators and other punctuation) not handled when converting variable names to canonical IDs
I received access to the model referenced in discussion #790.  There are a few issues affecting that model that need to be fixed in SDEverywhere.  The following describes one of those defects.

## Current Behavior

The `canonicalId()` function in `packages/parse/src/_shared/canonical-id.js` converts Vensim variable names into C/JS identifiers. It handles a specific set of special characters (`'".,\-$&%/|()`) but does NOT handle other characters that Vensim allows in variable names, such as:

- Operators: `+`, `>`, `<`, `=`
- Other symbols: `#`, `@`, `^`, `*`, `~`, `?`, `:`, `;`

Any of those (and other) symbols should be valid as long as they are inside double quotes, per the Vensim docs (see [Rules for Variable Names](https://www.vensim.com/documentation/ref_variable_names.html)).

When a Vensim model contains a variable name like `"A+B"` or `"GDP>0"`, the generated identifier contains invalid characters (`_a+b`, `_gdp>0`) which cause JavaScript/C syntax errors.

Real-world models contain variable names with `+`, `>`, `<`, and accented characters. Users currently have to apply post-processing workarounds to fix the generated code.

## Expected Behavior

We should change the regex so that other special characters are converted to underscore like we do for the current limited set of special characters.

## Notes

* There is a known limitation of our current approach for converting to canonical IDs: it's possible for two different variable names to map to the same ID, for example `"x > y"` and `"x < y"` would both map to `__x___y`.  We should consider a different strategy for converting names to IDs to avoid this, but it's not likely to come up in practice so I will leave that for another day (it will not be addressed as part of this issue).

chrispcampbell:
Fixes #794 

See issue for details.  Instead of handling specific characters inside a quoted variable name, we now convert anything that's not a letter or digit to an underscore.  I tested with all existing sample models, with En-ROADS and C-ROADS, and with the model from the referenced discussion.

### AI disclosure

I guided Claude Code (Opus 4.6) to investigate the issue and implement the changes.  I improved the tests that Claude wrote, updated the comments, and reviewed the changes.

### DIFF
--- a/packages/parse/src/_shared/canonical-id.js
+++ b/packages/parse/src/_shared/canonical-id.js
@@ -6,10 +6,14 @@ const reTrailingMark = new RegExp('\\s+!$', 'g')
 // Detect one or more consecutive whitespace or underscore characters
 const reWhitespace = new RegExp('(\\s|_)+', 'g')
 
-// Detect special punctuation characters
-// TODO: We do not currently include '!' characters in this set; we should only replace these
-// when they don't appear at the end of a (marked) dimension
-const reSpecialChars = new RegExp(`['"\\.,\\-\\$&%\\/\\|()]`, 'g')
+// Detect characters that are not valid in C/JS identifiers.  This matches anything
+// that is not a Unicode letter (\p{L}), Unicode digit (\p{N}), underscore,
+// or '!' (which is preserved for dimension marks).
+// TODO: Note that this means that we don't currently replace '!' characters that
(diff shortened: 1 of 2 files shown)


ToddFincannonEI (approved):
This will always produce valid C identifiers, so it looks good to me.

PR #799, opened 2026-04-09
fix: correct handling of blank and non-numeric cells for `GET DIRECT DATA` and `GET DIRECT LOOKUPS`

### LINKED ISSUE #796: Handling of non-numeric cells differs from Vensim for `GET DIRECT DATA` and `GET DIRECT LOOKUPS`
I received access to the model referenced in discussion #790. There are a few issues affecting that model that need to be fixed in SDEverywhere. The following describes one of those defects.

## Current Behavior

The SDE compiler does not handle non-numeric / blank cells consistently for the `GET DIRECT DATA` and `GET DIRECT LOOKUPS` functions, and its behavior differs from that of Vensim.

This is observed in the model referenced in discussion #790.  That model uses a number of complex XLSX files.  In some of those files, there are multiple tables pasted into the same spreadsheet, often occupying the same rows as other tables.  Vensim will correctly stop reading when it reaches the first non-numeric or blank time value, but SDE keeps going and this can cause problems.

Here's a screenshot of a simplified Excel file I created that demonstrates the case where SDE breaks down:
Image

If we have the following Vensim equation:

```
x from csv = GET DIRECT DATA('data/data.csv', ',', '2', 'D3')
```

It should only read the 3 values from D3, E3, F3, but currently SDE will keep looking at row 2 and will log errors when it gets to the time values in H2, etc.  Then because non-numeric values are not currently handled well in general, there are also hard errors that stop compilation.

Here's what gets printed to the console if trying to run the `directdata_endofdata` sample model that I created for this issue:

```
Testing the directdata_endofdata model
ERROR: cannot convert "ignored 1" to a number in /Users/campbell/Projects/sdeverywhere1/models/directdata_endofdata/data/data.csv
ERROR: cannot convert "ignored 2" to a number in /Users/campbell/Projects/sdeverywhere1/models/directdata_endofdata/data/data.csv
ERROR: cannot convert "material" to a number in /Users/campbell/Projects/sdeverywhere1/models/directdata_endofdata/data/data.csv
ERROR: cannot convert "data" to a number in /Users/campbell/Projects/sdeverywhere1/models/directdata_endofdata/data/data.csv
ERROR: cannot convert "text" to a number in /Users/campbell/Projects/sdeverywhere1/models/directdata_endofdata/data/data.csv
ERROR: cannot convert "text" to a number in /Users/campbell/Projects/sdeverywhere1/models/directdata_endofdata/data/data.csv
ERROR: cannot convert "text" to a number in /Users/campbell/Projects/sdeverywhere1/models/directdata_endofdata/data/data.csv
Error: ERROR: cannot convert "text" to a number
    at throwError (file:///Users/campbell/Projects/sdeverywhere1/packages/compile/src/_shared/helpers.js:133:11)
    at cdbl (file:///Users/campbell/Projects/sdeverywhere1/packages/compile/src/_shared/helpers.js:144:9)
    at file:///Users/campbell/Projects/sdeverywhere1/packages/compile/src/generate/direct-data-helpers.js:51:31
    at generateDirectDataLookup (file:///Users/campbell/Projects/sdeverywhere1/packages/compile/src/generate/gen-lookup-from-direct.js:98:17)
    at generateLookupsFromDirectData (file:///Users/campbell/Projects/sdeverywhere1/packages/compile/src/generate/gen-lookup-from-direct.js:56:11)
    at generateEquation (file:///Users/campbell/Projects/sdeverywhere1/packages/compile/src/generate/gen-equation.js:105:14)
    at file:///Users/campbell/Projects/sdeverywhere1/packages/compile/src/generate/gen-code-c.js:22:12
    at _map (/Users/campbell/Projects/sdeverywhere1/node_modules/.pnpm/ramda@0.27.2/node_modules/ramda/src/internal/_map.js:7:19)
    at map (/Users/campbell/Projects/sdeverywhere1/node_modules/.pnpm/ramda@0.27.2/node_modules/ramda/src/map.js:83:14)
    at /Users/campbell/Projects/sdeverywhere1/node_modules/.pnpm/ramda@0.27.2/node_modules/ramda/src/internal/_dispatchable.js:50:15
```

## Expected Behavior

SDE should match Vensim behavior as best as possible for these edge cases:
* We should stop processing data points when we either a) hit a non-numeric / blank cell for the time value or b) reach the end of the row/column.
* We should silently skip data points (i.e., omit them from the generated lookup) if the cell for the value is non-numeric or blank.

chrispcampbell:
Fixes #796 

See issue for details.  In addition to the new sample models, I tested this with En-ROADS and the model in the referenced discussion.

### AI disclosure

I guided Claude Code (Opus 4.6) to investigate the issue and implement the initial round of changes.  I manually wrote the sample mdl files (and associated xlsx/csv files) and ran them in Vensim to figure out the expected behavior.  I wasn't happy with Claude's first couple iterations on the fixes, so I guided Claude using the sample data files and refined the code manually.


ToddFincannonEI (approved):
It makes sense to stop reading data in this condition. In EPS we found it unacceptable to return zero for invalid cells, and so we have a separate data check in the build process before code gen. I agree that duplicating Vensim behavior is correct for SDEverywhere.

PR #802, opened 2026-04-10
fix: visit variable ref in lookup-call expressions to ensure that marked dimensions are detected

### LINKED ISSUE #801: Loop code not emitted when lookup call is used inside SUM expression
I received access to the model referenced in discussion #790.  There are a few issues affecting that model that need to be fixed in SDEverywhere.  The following describes one of those defects.

## Current Behavior

When a `SUM` expression contains a lookup call on a subscripted variable (e.g., `SUM(LookupVar[Dim!](Time))`), the compiler fails to generate the required `for` loop over the marked dimension. The loop variable (e.g., `u`) appears in the generated code but is never declared, causing a runtime `ReferenceError`.

This does not affect `SUM` calls over plain variable references (e.g., `SUM(x[Dim!])`), which work correctly.

Here's a simple Vensim model that can be used to demonstrate the problem:

```
      DimA: A1, A2 ~~|
      x[A1]( [(0,0)-(2,2)], (0,0),(2,1.3) ) ~~|
      x[A2]( [(0,0)-(2,2)], (0,0.5),(2,1.5) ) ~~|
      y = SUM(x[DimA!](1)) ~~|
```

Currently the compiler generates the following incorrect JS code for `y`:

```js
let __t1 = 0.0;
__t1 += fns.LOOKUP(_x[u], 1.0);
_y = __t1;
```

## Expected Behavior

The loop code should be emitted:

```js
let __t1 = 0.0;
for (let u = 0; u < 2; u++) {
__t1 += fns.LOOKUP(_x[u], 1.0);
}
_y = __t1;
```

chrispcampbell:
Fixes #801 

@ToddFincannonEI: One more PR for you.  (And thanks for the other reviews you just completed.)  See issue for details.  This one is pretty straightforward with a simple fix and added tests.


ToddFincannonEI (approved):
good catch, straightforward fix

PR #804, opened 2026-04-13
fix: add --datadir option to `sde generate` to allow for resolving data file paths in their original location

### LINKED ISSUE #303: External data files not resolved when preprocessed model placed in `sde-prep` directory
Currently the `build` package uses the `sde-prep` directory as a "working directory" and writes the preprocessed mdl file (the output of `sde generate --preprocess`) to that directory.  If the mdl file directly references other external data files, then these file references won't work because the mdl file is now at a different level than the data files.  For example:

The following is an example of an mdl file and a project directory structure that doesn't currently work:
```
X = GET DIRECT CONSTANTS('data.csv', ...)
```

```
project/
  model-files/
    data.csv
    model.mdl
  sde-prep/
    processed.mdl
```

The `ext-control-params` integration test is affected by this issue.

The workaround for now is to use one of the config file callbacks to edit the paths so that `data.csv` is replaced with `../model-files/data.csv`.

chrispcampbell:
Fixes #303 

@ToddFincannonEI: See issue for details.  This issue has required workarounds to be added in most every user-submitted project I've seen, so it will be nice to not have to use those workarounds any longer.  This was also one of the issues on the "1.0" milestone list, so it will be good to have this one checked off.

There is no impact on existing users that use `sde generate` directly, since the data files will continue to be resolved relative to the original mdl file's directory by default.

I think ideally the `modelDir[name]` parameter (that gets passed through in a few different places in the compile package) would ideally be called `dataDir` since it isn't actually used to read the mdl file, but changing that would've touched many more lines unnecessarily, so I left it as is and only updated a couple `@param` docs for clarity.

### AI disclosure

I used Claude Code (Opus 4.6) to check my proposed solution, but it didn't come up with any better alternatives.  I manually wrote the new integration test.


ToddFincannonEI (approved):
EPS uses sde generate directly, so this will not apply to us. I will leave this fix to your ample good judgment.

PR #805, opened 2026-04-13
fix: correct generated code for apply-to-all lookup definitions

### LINKED ISSUE #642: Incorrect code generated for apply-to-all lookup definitions
Currently the transpiler will generate incorrect code for an "apply-to-all" lookup definition, that is, a lookup definition where the lookup data is the same for all subscripts in a dimension (aka subscript range).  This is not a common case because it is not too useful to define a lookup like this (since you can get the same effect by defining the lookup without the dimension(s) on the LHS), but it sometimes comes up in practice if the lookup is added as a placeholder until subscript-specific lookups are defined at a later time.

For example:
```
DimA: A1, A1 ~~|
x[DimA]( (0,10), (1,20) ) ~~|
y[DimA] = x[DimA](1) ~~|
```

Currently the transpiler incorrectly generates a useless for loop at decl time:

```js
for (let i = 0; i < 2; i++) {
  const _x_data__i_ = [0.0, 10.0, 1.0, 20.0];
}
```

And the init code references `_x_data__i_` in a loop, which is fine, but I think the variable name could be simplified to just `_x_data`:

```js
for (let i = 0; i < 2; i++) {
  _x[i] = fns.createLookup(2, _x_data__i_);
}
```

---

### Workaround

The workaround is to define the lookup without dimensions:
```
DimA: A1, A1 ~~|
x( (0,10), (1,20) ) ~~|
y[DimA] = x(1) ~~|
```

---

### Test case

This is not a high priority issue since there is a simple workaround, but we should fix this eventually since there's no reason that the transpiler should fail in this case (and I'd say it's OK if the generated code is not too efficient for this case).

Here is a failing unit test that I've added to a branch to be addressed eventually:
```ts
  it('should work for lookup definition (1D, apply-to-all)', () => {
    const vars = readInlineModel(`
      DimA: A1, A2 ~~|
      x[DimA]( (0,10), (1,20) ) ~~|
    `)
    expect(vars.size).toBe(1)
    expect(genJS(vars.get('_x'), 'decl')).toEqual(['const _x_data = [0.0, 10.0, 1.0, 20.0];'])
    expect(genJS(vars.get('_x'), 'init-lookups')).toEqual([
      'for (let i = 0; i < 2; i++) {',
      '_x[i] = fns.createLookup(2, _x_data);',
      '}'
    ])
  })
```

chrispcampbell:
Fixes #642 

@ToddFincannonEI: This is the second of three today.  (I have a few more perf fixes on the way, but I will save those for later in the week.)  See issue for details.

### AI disclosure

I manually wrote the test cases first.  I then used Claude Code (Opus 4.6) to investigate a fix.  Its first attempt was more complicated than it needed to be.  I eventually simplified it, recognizing that the existing variable names were fine and that we just needed to omit the for loops in `decl` mode.


ToddFincannonEI (approved):
Good to fix this up for 1.0, even if we can attribute it to user error. It made me wonder whether we're doing this in EPS, which is so heavily subscripted, but fortunately we are not.

PR #809, opened 2026-04-20
feat: add support for Vensim's ALLOCATE BY PRIORITY function

### LINKED ISSUE #792: Implement Vensim's ALLOCATE BY PRIORITY
Add support for [Allocate by Priority](https://www.vensim.com/documentation/fn_allocate_by_priority.html) for generated C code.
This function distributes a limited quantity across multiple demands based on priority order. Higher-priority demands are satisfied first, and any remaining quantity is allocated to lower-priority demands.

Note: I have started working on this feature. Will raise a PR for it in a couple of weeks. I am only planning on adding support for C initially, but can work on the JS portion slowly too.

NiharikaHari:
Fixes #792 

Feature: Allocate by priority #792 

- Adds support for Vensim’s allocate by priority function in the C runtime.
- Mirrors the structure and style of the existing `allocate_available` implementation for consistency.
- Includes Vensim’s example model (with .dat file) under models/ for validation and comparison of outputs.
- **Note:** JS support is not implemented yet.


chrispcampbell:
@NiharikaHari: Thank you for your contribution!  I have some high priority tasks to attend to for the next couple days, but I will try my best to review your work by Thursday.

packages/cli/src/c/vensim.c
```diff
@@ -503,6 +503,234 @@ double* _ALLOCATE_AVAILABLE(
...
+  }
+
+  // If supply = 0, all targets get allocated 0
+  if(fabs(supply) < _epsilon) {
+    return allocations;
+  }
+
+  static double out_return[ALLOCATE_BY_PRIORITY_BUFSIZE];
+
+  fprintf(stderr, request_quantities);
```
chrispcampbell: We generally don't leave printf's in production code.  I know that Todd included some printf's in the `_ALLOCATE_AVAILABLE` implementation above, which was fine because they are gated by `#ifdef PRINT_ALLOCATIONS_DEBUG_INFO`.  So it's OK if you want to leave these `fprintf` calls here, but just make sure to wrap them in `#ifdef PRINT_ALLOCATIONS_DEBUG_INFO` / `#endif` blocks.

packages/cli/src/c/vensim.c
```diff
@@ -503,6 +503,234 @@ double* _ALLOCATE_AVAILABLE(
...
+
+  }
+
+  // Return the distributed supply in the original order
+  // adding to it again the request 0 if the where removed
+  for (size_t i = 0; i < m; i++) {
+    out_return[idx[i]] = out[i];
+  }
+
+  fprintf(stderr, out_return);
```
chrispcampbell: Same comment as above about removing this or putting it in an `#ifdef` block.

packages/cli/src/c/vensim.c
```diff
@@ -503,6 +503,234 @@ double* _ALLOCATE_AVAILABLE(
...
+  return out_return;
+
```
chrispcampbell: Remove these two blank lines.

packages/cli/src/c/vensim.c
```diff
@@ -503,6 +503,234 @@ double* _ALLOCATE_AVAILABLE(
...
+      for (size_t i = 0; i < m; i++) {
+        if (fabs(out[i] - request[i]) <= 1e-12) {
+          active[i] = false;
+        }
+      }
+    }
+
+    // Reduce remaining supply
+    supply -= dx;
+
```
chrispcampbell: Remove blank line.

packages/cli/src/c/vensim.c
```diff
@@ -503,6 +503,234 @@ double* _ALLOCATE_AVAILABLE(
...
+  double request[ALLOCATE_BY_PRIORITY_BUFSIZE];
+  double priority[ALLOCATE_BY_PRIORITY_BUFSIZE];
+
+  for (size_t i = 0; i < m; i++) {
+    request[i] = (double)request_quantities[idx[i]];
+    priority[i] = priority_values[idx[i]];
+  }
+
+  // Create the outputs array
+  for (size_t i = 0; i < num_requesters; i++) out_return[i] = 0.0;
```
chrispcampbell: I'm not a fan of single-line loops or conditionals.  Make this:

```
for (size_t i = 0; i < num_requesters; i++) {
  out_return[i] = 0.0;
}
```

packages/cli/src/c/vensim.c
```diff
@@ -503,6 +503,234 @@ double* _ALLOCATE_AVAILABLE(
...
+
+  double out[ALLOCATE_BY_PRIORITY_BUFSIZE] = {0.0};
+
+  // Compute the distances between target supply and next target start
+  double distances[ALLOCATE_BY_PRIORITY_BUFSIZE];
+
+  for (size_t i = 0; i < m; i++) distances[i] = NAN;
```
chrispcampbell: Same comment about avoiding single-line loops.

packages/cli/src/c/vensim.c
```diff
@@ -503,6 +503,234 @@ double* _ALLOCATE_AVAILABLE(
...
+  // Index of the current activated target
+  bool active[ALLOCATE_BY_PRIORITY_BUFSIZE] = {false};
+  active[0] = true;
+
+  // Index of the last activated target
+  size_t c_i = 0;
+
+  // Continue allocating until supply is exhausted
+  while (supply > _epsilon) {
+
```
chrispcampbell: Remove blank line.

packages/cli/src/c/vensim.c
```diff
@@ -503,6 +503,234 @@ double* _ALLOCATE_AVAILABLE(
...
+    // Check if there are any active targets left
+    bool any_active = false;
+    for (size_t i = 0; i < m; i++) {
+      if (active[i]) {
+        any_active = true;
+        break;
+      }
+    }
+    if (!any_active) break;
```
chrispcampbell: Convert single-line conditional to multiple lines with braces.

packages/cli/src/c/vensim.c
```diff
@@ -503,6 +503,234 @@ double* _ALLOCATE_AVAILABLE(
...
+
+    // Compute how much supply is needed to activate the next target
+    // (last target will return nan)
+    double dx_next_start = (distances[c_i] - out[c_i]) / slopes[c_i];
+
+    // Determine next allocation step size:
+    // smallest of (next completion, next activation, remaining supply)
+    double dx = dx_next_top;
+
+    if (!isnan(dx_next_start) && dx_next_start < dx) dx = dx_next_start;
```
chrispcampbell: Convert single-line conditional to multiple lines with braces.

packages/cli/src/c/vensim.c
```diff
@@ -503,6 +503,234 @@ double* _ALLOCATE_AVAILABLE(
...
+    if (supply < dx) dx = supply;
```
chrispcampbell: Convert single-line conditional to multiple lines with braces.

packages/compile/src/model/read-equations.js
```diff
@@ -880,6 +885,9 @@ function visitFunctionCall(v, callExpr, context) {
           }
         }
         continue
+      } else if ((callExpr.fnId === '_ALLOCATE_BY_PRIORITY') & (index < 2)) {
```
chrispcampbell: There is a typo here: `&` should be `&&`.

Once you fix that, you should be able to remove the extra parentheses as they aren't needed.

Actually, I think you could probably just remove the `index < 2` part here.  Ideally we would have validation code here for each argument (similar to what the TODO comments say in the code that handles `ALLOCATE AVAILABLE` args above).  But if we're not going to add that validation code now, I think you can drop the `index < 2` part, otherwise it gives the impression that only those args need to be considered when really all args should get special treatment eventually.

chrispcampbell (changes requested):
Thank you @NiharikaHari for taking the time to contribute your work on this new function!

Most of my comments below are nitpicky things about code style.  I haven't read the logic too closely for the `_ALLOCATE_BY_PRIORITY` C runtime function, but as long as it produces equivalent results to Vensim (and I can see that it does since the tests are passing), then I will trust that the logic works.

The `test_js` job is failing; that job does more than just run JS-related tests, it also runs the JS-level integration tests that exercise both C/Wasm and JS code gen, and there are Emscripten warnings flagged here (see [build log](https://github.com/climateinteractive/SDEverywhere/actions/runs/24653976956/job/72803543818#step:12:240)) that will need to be resolved.

Also, please add at least one new test in `gen-equation-c-from-vensim.spec.ts` to verify that code gen works as expected.  You can follow the patterns used in the existing `should work for ALLOCATE AVAILABLE function (1D LHS, ...)` test.  Add your new test just after the last `ALLOCATE AVAILABLE` test in that file.

NiharikaHari:
Thank you @chrispcampbell for taking the time to review this. I'll be able to work on these changes next week. The logic in `_ALLOCATE_BY_PRIORITY` C runtime function mirrors the algorithm used in [PySD](https://github.com/SDXorg/pysd).

travisfranck:
@NiharikaHari This is a useful addition to SDE.  Thank you for contributing!  Does you model run now?

NiharikaHari:
@travisfranck Yes, it does! Outside of this I'm doing some janky changes around `setConstants` and `setLookups` to read from CIN files and CSV respectively in C runtime. Excited to integrate SDE into our pipeline :)

packages/compile/src/model/read-equations.js
NiharikaHari: Makes sense, will remove index < 2 part for now.

NiharikaHari:
@chrispcampbell I have made the requested changes. The failures in `test_js` look like they were because of the printf statements, which I have removed now.

travisfranck:
> @travisfranck Yes, it does! Outside of this I'm doing some janky changes around `setConstants` and `setLookups` to read from CIN files and CSV respectively in C runtime. Excited to integrate SDE into our pipeline :)

Congrats!  Again, thank you for contributing to SDE.

chrispcampbell:
@NiharikaHari:

> The failures in `test_js` look like they were because of the printf statements, which I have removed now.

You're correct that removing the printf statements resolved the first set of errors, but there was one more at the bottom, so the build was still failing:
image

We need to update the `modeltests` script so that it skips your new model when running tests for the JS target, just like we did for the `allocate` model.  I made the change already and pushed to your branch.

chrispcampbell (approved):
I verified that you addressed all of the review comments.  All good now, and the builds are successful.  I will merge shortly and will publish new versions of the `@sdeverywhere` packages in a bit.  Thanks again for contributing!

PR #833, opened 2026-06-03
fix: replace SheetJS dependency with custom XLSX file reader

### LINKED ISSUE #832: Replace SheetJS (xlsx package) with custom XLSX file reader
The compile package depends on SheetJS (xlsx package), which is no longer distributed via npmjs and must be pulled from `https://cdn.sheetjs.com`.

In #463 (back in 2024) we migrated from the no-longer-supported npmjs-hosted package to the custom-hosted one.  At the time I noted that I wasn't a fan of that solution, but it was the easiest approach.

Recently in pnpm 10.34.0, there was a change to tarball checks that caused this package to be flagged.  We worked around this in #829, but it is a fragile solution and the fact that we use a custom-hosted tarball leads to downstream consequences for users (especially those using pnpm) who depend on SDE's compile package.

Also recently I have been investigating a model (the one noted in discussion #790) that makes heavy use of xlsx files, so a large portion of the time generating the model is spent in SheetJS code.  I suspected that we could speed up that code if we used a different library or rolled our own.

I had Claude investigate performance and compatibility of a couple alternatives to SheetJS.  Based on that, I also had Claude create a simple XLSX file reader that only contains the tiny sliver of functionality that is needed by SDE.

Long story short, the custom reader is one JS file with ~450 lines of code (much of it comments).  In a real world test (generating the model from #790), total compile time goes from 35.8 seconds to 29.6 seconds (about 17% faster).

Though I'd prefer to use an off-the-shelf library, the ones out there are larger, slower, require async, and most come with extra dependencies.  Since the custom reader is relatively small and self contained (and that there aren't likely to be lots of SD models out there relying on unusual XLSX files) I'm not too worried about us maintaining the code, so I think we should go for it.

chrispcampbell:
Fixes #832 

@ToddFincannonEI: See issue for details.  As noted, this was largely implemented by Claude but I reviewed the code and tests and it all seems pretty straightforward.  Normally I would avoid going with a custom solution but in this case it seemed warranted and the performance wins are a nice bonus.

I tested this branch with the model referenced in the issue and it produces identical values as the previous SheetJS-based reader (the generated C/JS code is identical), and our existing XLSX tests pass as well, so I feel like it should be good enough for SDE's purposes.

**AI disclosure:** I guided Claude Code (Opus 4.7) to plan and implement the changes, and then I reviewed and refined the changes.

### DIFF
--- a/packages/compile/package.json
+++ b/packages/compile/package.json
@@ -19,6 +19,7 @@
     "@sdeverywhere/parse": "^0.1.4",
     "byline": "^5.0.0",
     "csv-parse": "^5.3.3",
+    "fflate": "^0.8.3",
     "ramda": "^0.27.0",
     "strip-bom": "^5.0.0",
     "xlsx": "https://cdn.sheetjs.com/xlsx-0.20.2/xlsx-0.20.2.tgz"
--- a/packages/compile/package.json
+++ b/packages/compile/package.json
@@ -21,8 +21,7 @@
     "csv-parse": "^5.3.3",
     "fflate": "^0.8.3",
     "ramda": "^0.27.0",
-    "strip-bom": "^5.0.0",
-    "xlsx": "https://cdn.sheetjs.com/xlsx-0.20.2/xlsx-0.20.2.tgz"
+    "strip-bom": "^5.0.0"
   },
   "author": "Climate Interactive",
   "license": "MIT",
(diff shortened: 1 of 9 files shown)


ToddFincannonEI (approved):
I think this is a great improvement. I'm glad we can move on from the problematic SheetJS. Getting the big performance boost is a big plus! I can use this code for a number of my own internal utilities as well. This is something I would never have even considered writing before agentic coding. The `xlsx.js` code is short but gnarly.

chrispcampbell:
Thanks @ToddFincannonEI.  I agree with all of the above.  It's not something I would've attempted before, and yeah, a bit gnarly but thankfully pretty terse.

PR #838, opened 2026-08-07
fix: correct ALLOCATE AVAILABLE convergence and code generation

### LINKED ISSUE #837: ALLOCATE AVAILABLE fails to converge on large allocations and restricts the avail argument
Three related problems with `ALLOCATE AVAILABLE` support, found while
building the Energy Policy Simulator (EPS) with SDEverywhere:

1. `_ALLOCATE_AVAILABLE` terminates its search using an absolute
   epsilon (`1e-6`) difference between total allocations and the
   available resource. When the allocated quantities are large (or very
   small), the search fails to converge (or converges prematurely),
   producing results that do not match Vensim.
2. The code generator requires the `avail` argument to be a variable
   reference, but Vensim accepts any expression there (e.g.
   `total supply * fraction`).
3. `ALLOCATIONS_BUFSIZE` is 60, which is too small for the dimensions
   used in the EPS.

There is also a small latent bug in the codegen helper
`cVarRefWithoutLastIndices`, which compares the subscripts array itself
(rather than its length) against the required count, so the validation
never fires.

A PR with fixes (and a small refactor that gives the allocation
functions their own C file) is ready to submit.

ToddFincannonEI:
Fixes #837

**Branch:** `todd/837-allocate-available-fixes` (base: `main`)

## Motivation

Energy Innovation uses SDEverywhere to build the Energy Policy Simulator
(EPS), which makes heavy use of `ALLOCATE AVAILABLE` over large
dimensions. This PR collects the fixes we needed to match Vensim's
results and handle EPS-scale models, plus a refactor that gives the
growing family of allocation functions its own C file.

## Changes

1. **Convergence fix**: `_ALLOCATE_AVAILABLE` terminated its binary
   search using an absolute epsilon difference, which fails to converge
   (or converges prematurely) when total allocations are large or tiny.
   It now uses a relative difference, falling back to absolute only when
   either value is near zero.
2. **Codegen fixes**: the `avail` argument of `ALLOCATE AVAILABLE` can
   now be any expression (Vensim allows e.g. `total supply * 0.5`), not
   just a variable reference. Also fixes the subscript-count validation
   in `cVarRefWithoutLastIndices`, which compared the subscripts array
   itself (rather than its length) against the required count.
3. **Buffer size**: `ALLOCATIONS_BUFSIZE` raised from 60 to 80 to
   accommodate the EPS's larger dimensions.
4. **Refactor**: `_ALLOCATE_AVAILABLE` (and its normal-distribution
   helpers) plus `_ALLOCATE_BY_PRIORITY` and its `__sum` helper (added
   recently in #809) move from `vensim.c` into a new `allocation.c`,
   in preparation for the additional allocation functions in the
   follow-on PR. Redundant `extern` declarations already provided by
   `sde.h` are removed from `vensim.c` and `macros.c`.

   One non-move change to flag for reviewers: #809 introduced an
   unconditional `#define PRINT_ALLOCATIONS_DEBUG_INFO` that appears to
   be vestigial (no code between it and the end of `vensim.c` used it).
   It is commented out in the move; leaving it active would have turned
   on debug printing for the allocation functions that now follow it in
   `allocation.c`.

## Testing

- `tests/modeltests allocate` and `tests/modeltests allocate_by_priority`
  pass (C target).
- `packages/compile` unit tests pass (833 tests).
- Full workspace build succeeds.

## Notes

This is the first of a three-PR stack:
1. this PR (fixes + refactor)
2. priority curve types + allocation by price
3. A&S CDF coefficient precision


chrispcampbell:
It is difficult to see the changes that were made to the C `ALLOCATE` functions because this PR includes both a refactoring (moving the code to a separate file) and fixes to that code.  I would've preferred having a separate PR for the refactoring (either before or after the PR for the fixes), but too late for that now.  Fortunately it's not too hard to apply the changes from the new `allocate.c` back to the original part of `vensim.c` to produce the diff below, which makes it easier to see the actual changes/fixes.

```diff
diff --git a/packages/cli/src/c/vensim.c b/packages/cli/src/c/vensim.c
index 9899179b..b946d95b 100644
--- a/packages/cli/src/c/vensim.c
+++ b/packages/cli/src/c/vensim.c
@@ -370,10 +370,32 @@ double* _VECTOR_SORT_ORDER(double* vector, size_t size, double direction) {
   return result;
 }

-//
-// ALLOCATE AVAILABLE
-//
-// Mathematical functions for calculating the normal pdf and cdf at a point x
+// ALLOCATE AVAILABLE distributes a resource among requesters using a priority
+// profile for each requester. The priority profile specifies a complementary
+// cumulative distribution function based on the normal distribution. The shape
+// of the distribution is given by the priority (indicating the midpoint) and
+// the width (spread). The search space for allocations that match the
+// available resource is the x axis. A greater priority pushes the midpoint of
+// the distribution to the right, resulting in more area under the curve at a
+// given x and a larger allocation for that requester.
+
+// Return true if the value is near zero.
+static inline bool __isZero(double value) { return fabs(value) < _epsilon; }
+// Compute the absolute difference when x or y is near zero, otherwise compute
+// the relative difference, with y considered as the baseline.
+static inline double __difference(double x, double y) {
+  double diff = 0.0;
+  if (__isZero(x) || __isZero(y)) {
+    diff = fabs(x - y);
+  } else {
+    diff = fabs(1.0 - x / y);
+  }
+  return diff;
+}
+// Return true if the values are equal up to the tolerance.
+static inline bool __isEqual(double x, double y) { return __difference(x, y) < _epsilon; }
+
+// Normal distribution
 double __pdf_normal(double x, double mu, double sigma) {
   double base = 1.0 / (sigma * sqrt(2.0 * M_PI));
   double exponent = -pow(x - mu, 2.0) / (2.0 * sigma * sigma);
@@ -403,15 +425,12 @@ double __get_pp(double* pp, size_t iProfile, size_t iElement) {
   const int NUM_PP = PEXTRA - PTYPE + 1;
   return *(pp + iProfile * NUM_PP + iElement);
 }
-#define ALLOCATIONS_BUFSIZE 60
+#define ALLOCATIONS_BUFSIZE 80
 // #define PRINT_ALLOCATIONS_DEBUG_INFO
 double* _ALLOCATE_AVAILABLE(
     double* requested_quantities, double* priority_profiles, double available_resource, size_t num_requesters) {
   // requested_quantities points to an array of length num_requesters.
   // priority_profiles points to an array of num_requesters arrays of length 4.
-  // The priority profiles give the mean and standard deviation of normal curves used to allocate
-  // the available resource, with a higher mean indicating a higher priority. The search space for
-  // allocations that match the available resource is the x axis with tails on both ends of the curves.
   static double allocations[ALLOCATIONS_BUFSIZE];
   if (num_requesters > ALLOCATIONS_BUFSIZE) {
     fprintf(stderr, "_ALLOCATE_AVAILABLE num_requesters exceeds internal maximum size of %d\n", ALLOCATIONS_BUFSIZE);
@@ -440,7 +459,8 @@ double* _ALLOCATE_AVAILABLE(
     min_mean = fmin(__get_pp(priority_profiles, i, PPRIORITY), min_mean);
     max_mean = fmax(__get_pp(priority_profiles, i, PPRIORITY), max_mean);
   }
-  // Start the search in the midpoint of the means, with a big first jump scaled to the spread of the means.
+  // Start the search in the midpoint of the means, with a big first jump scaled
+  // to the spread of the means.
   double total_allocations = 0.0;
   double x = (max_mean + min_mean) / 2.0;
   double delta = (max_mean - min_mean) / 2.0;
@@ -469,8 +489,10 @@ double* _ALLOCATE_AVAILABLE(
       total_allocations += allocations[i];
     }
 #ifdef PRINT_ALLOCATIONS_DEBUG_INFO
-    fprintf(stderr, "x=%-+14g delta=%-+14g Δ=%-+14g total_allocations=%-+14g available_resource=%-+14g\n", x, delta,
-        fabs(total_allocations - available_resource), total_allocations, available_resource);
+    fprintf(stderr,
+        "x=%-+14g delta=%-+14g diff=%-14g%% total_allocations=%-+14g "
+        "available_resource=%-+14g\n",
+        x, delta, __difference(total_allocations, available_resource) * 100.0, total_allocations, available_resource);
 #endif
     if (++num_steps >= max_steps) {
       fprintf(stderr,
@@ -489,11 +511,12 @@ double* _ALLOCATE_AVAILABLE(
     last_delta_sign = delta_sign;
     delta = (delta_sign * fabs(delta)) / (num_jumps_in_same_direction < 3 ? 2.0 : 1.0);
     x += delta;
-    // The search terminates when the total allocations are equal to the available resource
-    // up to a very small epsilon difference.
-  } while (fabs(total_allocations - available_resource) > _epsilon);
+    // The search terminates when the total allocations are equal to the
+    // available resource up to the built-in tolerance.
+  } while (!__isEqual(total_allocations, available_resource));
 #ifdef PRINT_ALLOCATIONS_DEBUG_INFO
-  fprintf(stderr, "converged with Δ=%g in %zu steps\n", fabs(total_allocations - available_resource), num_steps);
+  fprintf(stderr, "converged with diff=%g%% in %zu steps\n",
+      __difference(total_allocations, available_resource) * 100.0, num_steps);
   fprintf(stderr, "total_allocations=%f, available_resource=%f\n", total_allocations, available_resource);
   for (size_t i = 0; i < num_requesters; i++) {
     fprintf(stderr, "[%2zu] %f\n", i, allocations[i]);
@@ -518,7 +541,7 @@ double __sum(double* arr, size_t n) {
 // ALLOCATE BY PRIORITY
 //
 #define ALLOCATE_BY_PRIORITY_BUFSIZE 60
-#define PRINT_ALLOCATIONS_DEBUG_INFO
+// #define PRINT_ALLOCATIONS_DEBUG_INFO

 double* _ALLOCATE_BY_PRIORITY(
   double* request_quantities, double* priority_values, double size, double width, double supply, size_t num_requesters) {
```

packages/compile/src/generate/gen-expr.js
```diff
@@ -896,7 +896,7 @@ function generateAllocateAvailableCall(callExpr, ctx) {
   function cVarRefWithoutLastIndices(arg, count) {
     const varRef = ctx.cVarRef(arg)
     const origIndexParts = Model.splitRefId(varRef).subscripts
-    if (origIndexParts < count) {
+    if (origIndexParts.length < count) {
```
chrispcampbell, line 899: We should have a test to verify this.  I'm adding one now in `gen-equation-c-from-vensim.spec.ts` and will push shortly.

packages/compile/src/generate/gen-expr.js
```diff
@@ -918,9 +918,8 @@ function generateAllocateAvailableCall(callExpr, ctx) {
   const ppArg = validateArg(1, 'pp')
   const ppRef = cVarRefWithoutLastIndices(ppArg, 2)
 
-  // Process the avail argument; include any subscripts
-  const availArg = validateArg(2, 'avail')
-  const availRef = ctx.cVarRef(availArg)
+  // Process the avail argument; include any subscripts. The avail arg can be any expression.
+  const availArg = generateExpr(callExpr.args[2], ctx)
```
chrispcampbell, line 923: We should have tests to verify this.  I'm adding tests now in `gen-equation-c-from-vensim.spec.ts` and will push shortly.

packages/cli/src/c/makefile
```diff
@@ -1,5 +1,5 @@
 # Do "export P={c-file-basename}" before running this makefile.
-OBJECTS=main.o vensim.o model.o macros.o
+OBJECTS=main.o vensim.o allocation.o model.o macros.o
```
chrispcampbell, line 2: We have two other places where we specify the list of files to be compiled, one is in `plugin-wasm`, and the other is in the docs for the `runtime` package.

I will fix those places to include `allocation.c` and will push to your branch shortly.

chrispcampbell (commented):
The code changes in the C runtime functions seem fine to me.  I've noted some other needed changes inline but have already fixed them myself and will push them shortly.  I will approve and merge once I verify the builds.

chrispcampbell (approved):
I'm good with these changes now that the other noted fixes have been added.  Will merge now.

PR #840, opened 2026-08-07
feat: support all ALLOCATE AVAILABLE priority curve types and add FIND MARKET PRICE / DEMAND AT PRICE / SUPPLY AT PRICE

### LINKED ISSUE #839: Support all ALLOCATE AVAILABLE priority curve types and the allocation-by-price functions
SDEverywhere's `ALLOCATE AVAILABLE` implementation supports only the
normal-distribution priority curve (the `ptype` element of the priority
profile is effectively ignored). Vensim also defines fixed,
rectangular, triangular, and exponential curve types.

Separately, Vensim's market-clearing allocation functions —
`FIND MARKET PRICE`, `DEMAND AT PRICE`, and `SUPPLY AT PRICE` — are not
supported at all.

The Energy Policy Simulator uses the full set: several curve types in
`ALLOCATE AVAILABLE`, and by-price allocation to balance energy supply
and demand across technologies.

A PR is ready that implements the remaining curve types (with a new
`allocate_ptype` test model) and the three by-price functions (with a
new `allocate_price` test model), both validated against Vensim output.
It builds on the `ALLOCATE AVAILABLE` fixes PR.

ToddFincannonEI:
**Branch:** `todd/839-allocate-price` (base: `todd/837-allocate-available-fixes` — stacked on PR #838)

Fixes #839 

## Motivation

The Energy Policy Simulator allocates energy demand across technologies
using Vensim's full allocation toolkit: `ALLOCATE AVAILABLE` with
several priority curve types, and market-clearing allocation via
`FIND MARKET PRICE` with `DEMAND AT PRICE` / `SUPPLY AT PRICE`.
SDEverywhere previously supported only the normal-distribution curve in
`ALLOCATE AVAILABLE` and none of the by-price functions.

## Changes

Commit 1 — **priority curve types**:
- `_ALLOCATE_AVAILABLE` now supports the fixed, rectangular, triangular,
  normal, and exponential ptypes (each curve implemented as a CDF with a
  complementary variant), selected per-profile via the `ptype` element.
- Two behavior fixes that came out of testing against Vensim: return
  zero allocations when the available resource is zero or negative, and
  fall back to fixed (proportional) allocation when all priorities are
  equal.
- New `allocate_ptype` test model iterates all five ptypes across four
  request/supply regimes and compares against Vensim output.

Commit 2 — **allocation by price**:
- New C implementations of `_FIND_MARKET_PRICE` (searches for the price
  that balances total demand and supply allocations), `_DEMAND_AT_PRICE`,
  and `_SUPPLY_AT_PRICE` (individual allocations at a given price).
- Code generation and equation-reader support for the three functions,
  including the priority-profile argument handling shared with
  `ALLOCATE AVAILABLE`, and demander/supplier counts that honor the
  Vensim convention of passing either a dimension or the first element
  of the array.
- New `allocate_price` test model compared against Vensim output.

`tests/modeltests` adds the two new models to the reduced-precision
(1e-2) list, matching the existing `allocate` entry; comparisons at
1e-4 are not attainable for the iterative allocation searches.

JS code generation for these functions is not yet implemented (same
status as `ALLOCATE AVAILABLE` on `main`); the JS target throws a
"not yet implemented" error and the models are skipped for JS in
`tests/modeltests` (already covered by the `allocate*` glob).

## Testing

- `tests/modeltests allocate`, `allocate_ptype`, `allocate_price`, and
  `allocate_by_priority` pass (C target).
- `packages/compile` unit tests pass (833 tests).

## Notes

Stacked on `todd/allocate-available-fixes` (PR 1). The A&S CDF
coefficient precision change follows in PR 3.


chrispcampbell (approved):
The changes look good to me.  I trust that the new code works given that the new sample models are producing the expected results.  I added new vitest tests to verify read/codegen for the new functions, to mirror what we've done in the past for the `ALLOCATE` and other functions.

PR #842, opened 2026-08-07
fix: round A&S CDF coefficients to 6 significant digits to best match Vensim

### LINKED ISSUE #841: Normal-curve allocation results differ from Vensim due to CDF coefficient precision
The normal CDF used by `ALLOCATE AVAILABLE` is computed with the
Abramowitz & Stegun 26.2.17 rational approximation. Vensim appears to
use the same approximation but with coefficients rounded to 6
significant digits; SDEverywhere uses the full-precision values.

The difference is tiny per evaluation, but allocation models that are
sensitive to the curve tails (such as the Energy Policy Simulator)
accumulate discrepancies from Vensim output that exceed the comparison
tolerance. Rounding the coefficients to match Vensim brings the
results into agreement. (The evidence for Vensim's rounding is
empirical — matching outputs across many allocation scenarios.)

ToddFincannonEI:
**Branch:** `todd/841-cdf-precision` (base: `todd/839-allocate-price` — stacked on PR #840)

Fixes #841 

## Motivation

The normal-curve allocation in `ALLOCATE AVAILABLE` computes the normal
CDF with the Abramowitz & Stegun 26.2.17 rational approximation
(Hastings 5-term). Vensim evidently uses the same approximation but
with coefficients rounded to 6 significant digits. Using the
full-precision coefficients produces allocation results that differ
from Vensim by more than the comparison tolerance in models that are
sensitive to the curve tails (as the Energy Policy Simulator is).

## Changes

- `__cdf_unit_normal` in `allocation.c` now uses the 6-significant-digit
  rounded coefficients (e.g. `0.231642` instead of `0.2316419`),
  matching Vensim's observed behavior.

## Testing

- `tests/modeltests allocate` and `allocate_price` pass (C target)
  against Vensim reference data.

## Notes

Stacked on `todd/allocate-price` (PR 2), because the coefficient change
applies to the restructured CDF code introduced there. Verify the
claim about Vensim's rounding against your own Vensim runs if you have
a convenient case; our evidence is empirical (matching outputs across
EPS allocation scenarios).


chrispcampbell (approved):
Looks fine to me.

PR #844, opened 2026-08-07
fix: detect false cyclic dependencies and break them by separating variables automatically

### LINKED ISSUE #843: False cyclic dependency errors for models that Vensim evaluates element by element
Vensim evaluates array equations element by element, so a model can be
valid in Vensim even when its equations are circular at the
whole-array level — for example, an iterative pass where element `i+1`
of an array depends on values computed from element `i`. SDEverywhere
sorts whole variables in dependency order, reports a cyclic dependency
for this pattern, and refuses to generate code.

The existing escape hatch is to hand-maintain `specialSeparationDims`
entries in the spec file, separating the affected variables so the
cycle disappears. For a model the size of the Energy Policy Simulator,
discovering which variables and which (sub)dimensions to separate is
slow, error-prone manual work that must be redone as the model evolves.

A PR is ready that detects these false cycles during `Model.read`,
identifies the variables and dimensions whose separation breaks the
cycle, and re-reads the model with those separations applied
automatically — as if they had been listed in the spec file. Truly
circular models still fail with the cycle reported.

ToddFincannonEI:
**Branch:** `todd/843-cyclic-deps` (base: `main`)

Fixes #843 

## Motivation

Vensim evaluates array equations element by element, so a model can be
perfectly valid in Vensim even when its equations look circular at the
whole-array level. A common pattern in the Energy Policy Simulator: an
"optimization pass" where element `i+1` of an array depends on results
computed from element `i` of the same array (via a different variable).
Vensim happily evaluates this in element order. SDEverywhere, which
sorts whole variables in dependency order, reports a cyclic dependency
and refuses to generate code.

Until now the workaround was to hand-maintain `specialSeparationDims`
entries in the spec file — separating a variable into per-index
instances removes the false cycle. For a model the size of the EPS,
discovering *which* variables and dimensions to separate is slow,
error-prone manual work that has to be redone as the model evolves.

## Changes

Commit 1 — **detect and break cyclic dependencies**: when dependency
sorting fails, the error now carries the cycle clusters and the
dependency edges. The cycle analysis identifies variables whose
separation (on specific dimensions) would break the cycle, adds them to
`specialSeparationDims`, and re-reads the model — as if the modeler had
listed them in the spec file by hand. Truly circular models (ones that
are cyclic even element by element) still fail with the cycle reported.
Set `SDE_PRINT_CYCLES=1` to print each detected cycle and the
separations chosen.

Commit 2 — **detect when the separation dimension must be a specific
subdimension**: extends the cycle analysis to cases where separating on
the variable's full dimension is not sufficient or not correct — the
EPS optimization-pass pattern above needs the variable separated on the
particular (sub)dimension that carries the element-order dependency.
The analysis inspects the subscripts on the cycle's edges to pick that
dimension.

Commit 3 — **move detection into `Model.read`**: the detection/repair
loop runs during the read phase rather than at code generation, so
variable listings (`sde generate --list` and friends) benefit from the
same treatment and all consumers see the post-separation model.

## Testing

- New unit test in `gen-code-c-from-vensim.spec.ts` exercises cycle
  detection and automatic separation.
- `packages/compile` unit tests pass (834 tests).
- `tests/modeltests` passes for both C and JS targets (notably
  `separateall` and `prune`, which exercise the separation machinery).
- The EPS itself builds and validates against Vensim with these changes
  (this is the feature's origin).


chrispcampbell (approved):
This is a big change, but since it only affects the case where a cycle is detected, I'm not too concerned about it having an impact on existing models.  I didn't closely review the individual steps of the new functions but I will trust that it works based on the passing tests.

I updated a couple comments to be JSDoc-style to make it easier to see the inputs and outputs.

I also did a light refactoring to move the new cycle analysis code in `model.js` to a separate `analyze-cycles.js` file to avoid making the already-too-large `model.js` file any larger.

I'm fine merging this; it can be refined later if needed.

PR #846, opened 2026-08-07
feat: add support for the INVERT MATRIX Vensim function in C and JS

### LINKED ISSUE #845: Support the INVERT MATRIX Vensim function
Vensim's `INVERT MATRIX(matrix, n)` function is not supported by
SDEverywhere for either the C or JS target.

The Energy Policy Simulator needs it to compute a total Leontief
inverse matrix from a 43×43 input-output table, but it is a
general-purpose Vensim function that other models are likely to use as
well.

A PR is ready that implements `INVERT MATRIX` for both targets
(Gauss-Jordan elimination with partial pivoting; singular matrices
yield `_NA_` values), along with two test models validated against
Vensim output: a small `invertmatrix` covering edge cases and an
`invertmatrixio` model exercising the realistic input-output-table
use case.

ToddFincannonEI:
**Branch:** `todd/845-invert-matrix` (base: `main`)

Fixes #845 

## Motivation

The Energy Policy Simulator computes a total Leontief inverse matrix
from an input-output table, which requires Vensim's
`INVERT MATRIX(matrix, n)` function. This PR implements it for both the
C and JS targets.

## Changes

- **C runtime**: `_INVERT_MATRIX` in `vensim.c` inverts an n×n matrix
  using Gauss-Jordan elimination with partial pivoting. The input is
  not modified; the result buffer is reused across calls and grown as
  needed. A singular matrix returns `_NA_` values (matching Vensim's
  behavior of not producing usable output rather than crashing).
- **JS runtime**: equivalent `INVERT_MATRIX` implementation in
  `js-model-functions.ts`.
- **Code generation**: `INVERT MATRIX` inverts the whole 2D argument at
  once, so the call is emitted once before the LHS subscript loops are
  opened and the per-element expression reads from the temporary. The
  equation reader validates the two-argument call (the second argument
  is typically `ELMCOUNT(dim)`).
- **Test models**:
  - `invertmatrix` — small matrices covering identity, permutation,
    singular, and general cases.
  - `invertmatrixio` — a realistic 43×43 total input-output table (the
    EPS use case) read via `GET DIRECT CONSTANTS` from CSV, with the
    Leontief computation `(I - TIOT)^-1`. The Vensim reference data is
    deliberately kept small (~19k lines) by running only two time steps;
    nothing in the model is time-dependent.

## Testing

- `tests/modeltests invertmatrix` and `invertmatrixio` pass for both C
  and JS targets against Vensim reference data.
- `packages/compile` (833) and `packages/runtime` (86) unit tests pass.


chrispcampbell (approved):
Looks good to me.  I added vitest tests to verify read and codegen for the new function.  All passing.

PR #849, opened 2026-08-07
fix: correct XLSX reader to normalize line endings + process cells with attributes

### LINKED ISSUE #847: XLSX reader drops cells with attributes on `<v>` and does not normalize line endings
The custom XLSX reader introduced in #833 (`packages/compile/src/_shared/xlsx.js`)
diverges from SheetJS (and from conformant XML parsing) in two ways that
silently corrupt cell data.

**1. Cells whose `<v>` tag has attributes are skipped entirely.**
`parseSheetXml` locates the cached value with `body.indexOf('<v>')`, which
fails when the tag carries attributes. Excel writes
`<v xml:space="preserve">` for formula-cached string results with leading
or trailing whitespace, for example:

```xml
<c r="P187" s="251" t="str"><f>IF(ISBLANK(INDEX(...)),"",INDEX(...))</f><v xml:space="preserve">**Description:** ...</v></c>
```

The reader skips the cell, so the workbook reports it as empty. In the
Energy Policy Simulator's WebAppData.xlsx this silently blanked 458 of
2053 policy-lever description cells. The same pattern can drop shared
string (`t="s"`), boolean (`t="b"`), and numeric cells.

**2. Line endings are not normalized.**
XML 1.0 §2.11 requires parsers to translate `\r\n` and lone `\r` to `\n`,
and SheetJS additionally normalizes after entity decoding, so `&#13;`
followed by a literal newline yields `"\n"`. The reader's `decodeXmlText`
leaves the `\r` in place, so the same cell reads back as `"...\r\n"` where
SheetJS returns `"...\n"`. Any consumer that hashes or compares cell text
sees a different string; in the EPS this changed a content-hashed i18n
string id, which would orphan the string's existing translations.

ToddFincannonEI:
**Branch:** `todd/847-xlsx-reader-value-parsing` (base: `main`)

Fixes #847

## Motivation

The custom XLSX reader added in #833 skips any cell whose `<v>` tag has
attributes, because `parseSheetXml` locates the cached value with
`body.indexOf('<v>')`. Excel writes `<v xml:space="preserve">` for
formula-cached string results (`t="str"`) with leading or trailing
whitespace, so those cells silently read back as empty. In the Energy
Policy Simulator's WebAppData.xlsx this blanked 458 of 2053 policy-lever
description cells.

Separately, `decodeXmlText` does not normalize line endings. XML 1.0
§2.11 requires `\r\n` and lone `\r` to become `\n`, and SheetJS also
normalizes after entity decoding (`&#13;` + literal newline → `"\n"`).
Without this, cell text differs from what SheetJS returned for the same
file, which breaks consumers that hash or compare strings (in the EPS,
a content-hashed i18n string id changed, orphaning its translations).

## Changes

- `parseSheetXml` extracts the cached value with a new `getVText` helper
  (`/<v\b[^>]*>([\s\S]*?)<\/v>/`) that tolerates attributes on the tag.
  The shared-string, formula-string, boolean, and numeric branches all
  use it; a cell with no `<v>` element (uncalculated formula) is still
  skipped as before.
- `decodeXmlText` now normalizes `\r\n` and lone `\r` to `\n` via a new
  `normalizeEol` helper, applied after entity decoding so `&#13;` is
  normalized too (matching XML parser and SheetJS behavior).

## Testing

- Two new cases in `xlsx.spec.ts`:
  - cells with `<v xml:space="preserve">` across `t="str"`, numeric,
    shared-string, and boolean types
  - CRLF normalization for raw `\r`, `&#13;` entities, and rich-text
    shared strings
- Full `src/_shared/xlsx.spec.ts` suite passes (31 tests).
- End-to-end: the EPS Australia build's generated config.json is
  structurally identical to the config produced by the SheetJS-based
  reader before #833, confirming both regressions are closed.

## Notes

Both fixes only add values that were previously dropped or repair text
fidelity; no previously-read value changes except the `\r` removal.


chrispcampbell (approved):
Looks good to me.

PR #894, opened 2026-08-26
perf: mark some C functions as `static inline` to improve performance

### LINKED ISSUE #893: Mark some C functions as `static inline` to improve performance
A long time ago we converted a number of C function to be simple macros, but we left a few as functions due to the fact that the arguments are used multiple times in the body and macro form would cause them to potentially be evaluated multiple times.

From a performance investigation of the SDE-generated En-ROADS model, it was noted (by Claude) that a few simple functions are being called hundreds of thousands (or millions) of times per model run:
* `_ZIDZ`
* `_XIDZ`
* `_RAMP`
* `_PULSE` (to a lesser extent)

From Claude's summary:

> For a large model like En-ROADS, `_ZIDZ`, `_XIDZ`, and `_RAMP` are called millions of
times per model run (measured: 1.70 M, 239 k, and 137 k calls per run respectively).
They were defined as out-of-line functions in `vensim.c`, which is a separate translation
unit from the generated model code, so:
> - every call was a real function call, and
> - `_epsilon` was a mutable global that had to be re-loaded from memory on each call, and
> - the compiler could not common-subexpression-eliminate repeated identical calls.

Converting these to `static inline` functions and making `_epsilon` a `#define` improves En-ROADS Wasm run time by 5-7% (with `-O2`) at the price of a 2.4% increase in size from the inlining.

chrispcampbell:
Fixes #893 

See issue for details.  This helps improve performance of large models like En-ROADS and has no impact on outputs of generated models.

### DIFF
--- a/packages/cli/src/c/allocation.c
+++ b/packages/cli/src/c/allocation.c
@@ -24,7 +24,7 @@
 // #define PRINT_ALLOCATIONS_DEBUG_INFO
 
 // Return true if the value is near zero up to the epsilon tolerance.
-static inline bool __isZero(double value) { return fabs(value) < _epsilon; }
+static inline bool __isZero(double value) { return fabs(value) < SDE_EPSILON; }
 // Compute the absolute difference when x or y is near zero, otherwise compute
 // the relative difference, with y considered as the baseline.
 static inline double __difference(double x, double y) {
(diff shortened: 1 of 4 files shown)


ToddFincannonEI (approved):
Great job on the profiling. I'm amazed there is such low-hanging fruit still available. I have no idea why we didn't define epsilon in the preprocessor to begin with. The implementation looks correct.

PR #898, opened 2026-08-26
perf: emit non-overridable constants as C `static const` values

### LINKED ISSUE #896: Emit non-overridable constants as C `static const` literals to improve performance
From a performance investigation of the SDE-generated En-ROADS model, it was noted (by Claude) that we have a number of true constants (i.e., values that are never overridden at run time, such as input variables) in the generated code that are currently mutable `double` values but could be initialized as `static const`, which would improve performance.

From Claude's summary:

> A model constant is currently emitted as a mutable global that is assigned in
`initConstants`:

```c
double _retirement_effect_exponent;
...
void initConstants0() {
  _retirement_effect_exponent = 0.5;
}
```

> The C compiler cannot see the value, so every use has to be compiled as a load of an unknown
quantity. Emitting it as a literal instead:

```c
static const double _retirement_effect_exponent = 0.5;
```

> lets the compiler fold the value into the expressions that use it. The large majority of the
measured win comes from one specific consequence: En-ROADS' learning curves are written as
`POWER(x, <named constant>)` where the exponents are 0.5, 1, 2, and 3. Once LLVM can see
those values it strength-reduces the `pow` calls into `sqrt` and multiplies. `pow` is by far
the most expensive primitive in the run (~16% of Wasm time before this change).

This change improves En-ROADS Wasm run time by ~7-9% at the price of a small < 2% increase in model size.

chrispcampbell:
Fixes #896 

See issue for details.  I left this as an opt-in optimization (enabled via non-public environment variable) since it can have a small effect on model outputs values, but I will probably eventually enable it for En-ROADS since it provides a decent performance improvement.

### DIFF
--- a/packages/compile/src/generate/gen-code-c-from-vensim.spec.ts
+++ b/packages/compile/src/generate/gen-code-c-from-vensim.spec.ts
@@ -11,6 +11,24 @@ import { parseInlineVensimModel } from '../_tests/test-support'
 import { generateCode } from './gen-code'
 import { generateC } from './gen-code-c'
 
+/**
+ * Run the given function with the environment variable that enables constant folding set,
+ * then restore the previous value.
+ */
+function withConstLiteralsEnabled(f: () => string): string {
+  const saved = process.env.SDE_NONPUBLIC_EMIT_CONST_LITERALS
+  process.env.SDE_NONPUBLIC_EMIT_CONST_LITERALS = '1'
+  try {
+    return f()
+  } finally {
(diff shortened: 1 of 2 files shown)


ToddFincannonEI (approved):
I took only a cursory look at the code to make sure that the input vars check would work for EPS. I won't be able to try it until I do the next update of SDEverywhere in the EPS builder in a couple of weeks. Based on the tests and the opt-in decision, I think this can go forward. Another nice and simple optimization with significant impact!

PR #899, opened 2026-08-26
fix: handle cycles when reducing variables

### LINKED ISSUE #897: Aggressive variable reduction fails on any model with a feedback loop
We have an optional code path (gated by a non-public environment variable) that allows for more aggressive constant folding during the model analysis phase (reducing known constant expressions to a single literal when possible).  It is not enabled by default because it can cause tiny differences in model outputs and may not provide significant performance wins for most models.

When I implemented it, I left a TODO comment about it not handling cyclic dependencies properly.  We should fix this so that the option is usable.

chrispcampbell:
Fixes #897 

See issue for details.  The fix is simple; just return early instead of throwing, which skips any optimization.  This case was only likely to come up in practice when using the non-public / opt-in "aggressive" mode, so it should have no impact on existing models with default compilation.

### DIFF
--- a/packages/compile/src/model/reduce-variables.js
+++ b/packages/compile/src/model/reduce-variables.js
@@ -49,14 +49,22 @@ export function reduceVariables(variables, inputVarIds, mode) {
       return
     }
 
-    // Add this variable to the set of active ones
-    // TODO: Allow cycle if the current LHS is referenced on the RHS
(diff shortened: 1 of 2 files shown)


ToddFincannonEI (approved):
This looks like simple cleanup to me and is well covered by the tests, so let's go ahead with it.

PR #901, opened 2026-08-27
perf: evaluate time-invariant aux variables once instead of on every time step

### LINKED ISSUE #900: Evaluate time-invariant aux variables once instead of on every time step
NOTE: I'm not planning to pursue this change because I think its complexity doesn't justify the modest performance gain (and it is a potential maintenance burden).  I'm filing this issue anyway (and an associated PR to be closed) as a record of the idea and the implementation details in case someone wants to consider this again in the future.

From a performance investigation of the SDE-generated En-ROADS model, it was noted (by Claude) that we have a number of aux variables that are computed on every time step even though they do not depend on the `_time` variable and therefore could be computed just once per run.

To make this work, the compiler would need to identify which variables are time-invariant (and can be computed once per run) and which ones are time-dependent (and therefore need to be computed on every time step, like in our normal `evalAux` functions).

This change potentially improves En-ROADS Wasm run time by ~4% with no significant impact on model size.

chrispcampbell:
Fixes #900 

See issue for details.  As noted in the issue, I am submitting this PR only to capture the state of the experimental branch, and I will close this after submitting and will close the issue as "will not fix".

### DIFF
--- a/NOTES.md
+++ b/NOTES.md
@@ -163,11 +163,11 @@ treated as time varying, i.e. the previous behavior). This follows the existing
 - En-ROADS Wasm (`emcc -O2`), model-check perf comparison against the `develop` baseline
   bundle (Apple M1 Max, Chrome), three "Run" passes:
 
-  | run |  develop |  current | % change |
-  | --- | -------: | -------: | -------: |
-  | 1   |     23.9 |     23.0 |    -3.6% |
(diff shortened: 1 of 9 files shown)


chrispcampbell:
This is an experimental PR only, will not be merged.  Closing.

